PR #11, opened 2022-11-01
feat: Import cms signals handler module in order to register receivers

andrey-canon:
## Description
When a course is deleted in the lms service the elasticsearch record is not deleted since that operation is performed by this method https://github.com/openedx/edx-platform/blob/open-release/maple.master/cms/djangoapps/contentstore/signals/handlers.py#L73 which is triggered by a signal however that receiver method is  not register because that belongs to the cms service. This change import the handler module in order to force the method registration.

### DIFF
--- a/eox_nelp/apps.py
+++ b/eox_nelp/apps.py
@@ -31,6 +31,13 @@ class EoxNelpConfig(AppConfig):
         }
     }
 
+    def ready(self):
+        """
+        Method to perform actions after apps registry is ended.
+        """
+        # This is required in order to register the receiver inside handlers module.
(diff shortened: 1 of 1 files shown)


johanv26 (commented):
This is mindblowing to me, so each particular import could be performed here in the `ready` method.
=)

andrey-canon:
> This is mindblowing to me, so each particular import could be performed here in the `ready` method. =)

@johanv26  you can add imports wherever you want the good practice is to set that at the top of the file, however I'm doing that inside the ready method to ensure that all the modules, lms and cms, are already loaded

johanv26 (approved):

PR #12, opened 2022-11-10
feat: Add middleware file 

andrey-canon:
## Description
Add middleware file that includes anew class that allows to extend the registration fields based on tenant settings

### DIFF
--- a/eox_nelp/middleware.py
+++ b/eox_nelp/middleware.py
@@ -1,0 +1,115 @@
+"""Middleware file.
+
+Required NELP middlewares that allow to customize edx-platform.
+
+classes:
+    ExtendedProfileFieldsMiddleware: Set extended_profile_fields in registration form.
+"""
+from eox_nelp.edxapp_wrapper.site_configuration import configuration_helpers
+from eox_nelp.edxapp_wrapper.user_authn import RegistrationFormFactory
+from eox_nelp.edxapp_wrapper.user_api import accounts
+
+
+# Copy default values from https://github.com/eduNEXT/edunext-platform/blob/ednx-release/mango.master.nelp
+# /openedx/core/djangoapps/user_authn/views/registration_form.py#L308
+DEFAULT_EXTRA_FIELDS = [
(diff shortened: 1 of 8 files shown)


eox_nelp/middleware.py
```diff
@@ -0,0 +1,115 @@
...
+            'sport': ['soccer', 'football', 'tennis'],
+            'movie': ['Harry Potter', 'Karate kid']
+        }
+    """
+
+    def __init__(self, get_response):
+        self.get_response = get_response
+
+    def __call__(self, request):
+        # The following line resets to the default values, to avoid tenants overlapping.
```
johanv26, line 70: What do you mean by the tenant overlapping. So it will always reset to this list?
https://github.com/eduNEXT/eox-nelp/pull/12/files#diff-66a15828789bf584965751e47deaa844139ec753886961cc4c1450c06e733c70R15

andrey-canon: An example, tenant A, with field movie, will add that field but if the next request comes from tenant B and that tenant doesn't have extra fields without that line the tenant B will show the movie field, so yes it will always reset the list to the original value

johanv26: That is interesting. so in that condition, I can't have tenants with fewer fields. But I think that that is not problem... who should remove REGISTER_FIELDS xD.

andrey-canon: This modifies EXTRA_FIELDS inside the class RegisterFormFactory,  that is a constant value therefore you will never have less fields. https://github.com/eduNEXT/eox-nelp/pull/12/files#diff-66a15828789bf584965751e47deaa844139ec753886961cc4c1450c06e733c70R15  previous line is a identical copy  of EXTRA_FIELDS you can find that values here https://github.com/eduNEXT/edunext-platform/blob/ednx-release/mango.master.nelp/openedx/core/djangoapps/user_authn/views/registration_form.py#L308

johanv26 (approved):

PR #13, opened 2022-11-25
feat: Add option to set extended profile field labels in different la…

andrey-canon:
## Description
This set field labels based on the `extended_profile_field_translations `, if the setting has not set or doesn't include the field name the label will be the field name 

## Testing instructions
1. set extended profile fields in your tenant config, example 
```
    "extended_profile_fields": [
        "hobby",
        "sport",
        "movie"
    ]
```
2. Set extended profile field translations, example
```
    "extended_profile_fields_translations": {
        "ar": {
            "hobby": "\u0647\u0648\u0627\u064a\u0629"
        }
    },
```
The `extended_profile_fields_translations `must be a dictionary with language code keys and every language code key is a dictionary that must include the field that will be translated where the key is the field name and the value is the translation

3. Go to /register

## Results in local
just the field hobby was set

#### EN
image

#### AR

image

### DIFF
--- a/eox_nelp/middleware.py
+++ b/eox_nelp/middleware.py
@@ -61,6 +61,18 @@ class ExtendedProfileFieldsMiddleware:
             'sport': ['soccer', 'football', 'tennis'],
             'movie': ['Harry Potter', 'Karate kid']
         }
+
+        # if you want to add translations
+
+        extended_profile_fields_translations = {
+            "ar": {
+                'hobby': 'هواية',
+            },
+            "es": {
+                'sport': 'Deporte',
+                'movie': 'Pelicula'
+            }
+        }
     """
 
     def __init__(self, get_response):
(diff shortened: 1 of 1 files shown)


eox_nelp/middleware.py
```diff
@@ -93,18 +110,19 @@ def __call__(self, request):
...
         this method generates a specific handler for each new field with its specific label.
 
         Args:
             field: String
+            label: String or None
         Returns:
             func: Handler method
         """
-        label = field.capitalize()
+        label = label.capitalize()
```
johanv26, line 125: @andrey-canon do you think that in this way we can avoid this commit ?
https://github.com/eduNEXT/edunext-platform/commit/acc844c166ff381c0b9c46a2dde2b1572d59664a

johanv26, line 125: If label is None ? this could happen??
image

andrey-canon: with the current implementation that can not happen however I will remove the default value since that comes from a previous implementation

andrey-canon: I don't think so

johanv26: I have a question, if you remove the default value. Why this method `_generate_handler `https://github.com/eduNEXT/eox-nelp/pull/13/files#diff-66a15828789bf584965751e47deaa844139ec753886961cc4c1450c06e733c70R113
uses two argument if only use the label. ?

johanv26 (approved):

PR #14, opened 2022-12-28
feat: add cache programs command using eox-tenant

johanv26:
# Description
- This PR started in the following discussion.
https://github.com/eduNEXT/edunext-platform/pull/720
- The command to load cache programs using eox-tenant and routes is:
`./manage.py lms eox_tenant_cache_programs`
- This moves the base command to this eox_tenant implementation from https://github.com/openedx/edx-platform/blob/open-release/maple.master/openedx/core/djangoapps/catalog/management/commands/cache_programs.py to this approach.
You can check the differences here:
https://www.diffchecker.com/JelU5kQJ/
- From now is necessary to use eox-tenant in order to use eox-nelp
## Test stage pic
https://drive.google.com/drive/folders/1shgl1cmtEkzWmSJGmjsJoeok05jfbqAM

### DIFF
--- a/eox_nelp/management/commands/eox_tenant_cache_programs.py
+++ b/eox_nelp/management/commands/eox_tenant_cache_programs.py
@@ -1,0 +1,133 @@
+""""Management command to add program info to the cache using eox-tenant."""
+
+
+import logging
+import sys
+
+from django.contrib.auth import get_user_model
+from django.contrib.sites.models import Site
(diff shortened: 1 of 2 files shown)


andrey-canon:
These changes depends on eox-tenant however that library  doesn't belong to the requirements, please add it

andrey-canon (approved):

PR #20, opened 2023-02-09
Jlc/lightway courses api

johanv26:
# Description
Remove extra files that were only copied from https://github.com/openedx/edx-platform/tree/master/lms/djangoapps/course_api
- This reduces the number of additional files and extra lines of python having the endpoint with the same behavior.
- This helps to reduce coverage.
- Remove not used endpoint https://github.com/eduNEXT/eox-nelp/compare/jlc/lightway-courses-api?expand=1#diff-dcdcc257ba000993140367b6dd38e4436dcf1593148dad2f009bc25c44eeca95L12.

The next step would be to use all the corresponding backends for this change.
https://github.com/eduNEXT/eox-nelp/compare/jlc/lightway-courses-api?expand=1#diff-511ef1ef75b9641934eedb06898b14fdab802b303292ce578924862995be38e4R16-R19

## keep function
Peek 2023-02-09 16-35


andrey-canon (approved):
LGTM

PR #24, opened 2023-02-20
feat: add experience ADR

andrey-canon:
## Description
This describes briefly models and APIs that are necessary to complete this requirements https://edunext.atlassian.net/browse/NELP-294

### DIFF
--- a/docs/decisions/0001-course-experience.rst
+++ b/docs/decisions/0001-course-experience.rst
@@ -1,0 +1,283 @@
+Courses units in LMS
+--------------
+
+Status
+======
+
+Accepted
+
+Context
+=======
+
+NELP business case requires multiples features that are not part
+of open-edx platform core, and it's necessary a new implementation
+that allows users to report and share their opinions about the
+course content and experience during the course.
+
+Decisions
+=========
+
+1. Create a course experience module.
+2. Create ``RateCourse`` model with the following fields.
+
+  * author => Foreign key to user model, the user who set their opinion.
+
+  * status => String field with choices: Liked, disliked and not-set.
+
+  * course_id => String field, course identifier to allow course rating.
+
+3. Create ``RateSection`` model with the following fields.
+
+  * author => Foreign key to user model, the user who set their opinion.
+
+  * status => String field with choices: Liked, disliked and not-set.
+
+  * item_id => String field, this is course subsection identifier.
+
+4. Create ``ReportCourse`` model with the following fields.
+
+  * author => Foreign key to user model, the user who set their opinion.
+
+  * reason => String field with choices:
+
+   - Sexual content
+
+   - Graphic violence.
+
+   - Hateful or abusive content.
+
+   - Copycat or impersonation.
+
+   - Other objection.
+
+  * course_id => String field, course identifier to allow course rating.
+
+5. Create ``ReportSection`` model with the following fields.
+
+  * author => Foreign key to user model, the user who set their opinion.
+
+  * reason => String field with choices:
+
+   - Sexual content
+
+   - Graphic violence.
+
+   - Hateful or abusive content.
+
+   - Copycat or impersonation.
+
+   - Other objection.
+
+  * item_id => String field, this is course subsection identifier.
+
+6. Set model constrains, each model should be unique per sub-sections and user or course and user.
+7. Create API views for each model.
+
+
+API Examples for sub-sections
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+  GET request
+
+  /eox-nelp/api/experience/v1/rate/sections/<usage-id>/
+
+  Response
+
+  .. code-block:: json
+
+      {
+        "data": {
+         "type": "rate-section"
+         "attributes": {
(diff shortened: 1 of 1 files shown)


docs/decisions/0001-course-experience.rst
```diff
@@ -0,0 +1,156 @@
...
+
+4. Set model constrains, ``Like`` and ``Report`` model should be unique per sub-sections and user.
+5. Create API views for each model.
+
+
+API Examples
+~~~~~~~~~~~~
+  GET request
+
+  /eox-nelp/api/experience/v1/<usage-id>/rate
```
johanv26, line 55: That has to be in that url  shape?.  `api/experience/v1`  or could be like other endpoints: `experience/v1/`: 
image

 by the way, mfe config API that is "new " has that shape. api/mfe_config/v1/

andrey-canon: @johanv26 please expose arguments

johanv26: no, I am only asking what is the consideration to choose that shape of the API URL?

andrey-canon: commonly the word `api` is used to make a difference between view endpoints and endpoints that returns just json data, in django is very common to have  both those views, so this is a good way to identify each one, however this is not an standard or hard rule

docs/decisions/0001-course-experience.rst
```diff
@@ -0,0 +1,283 @@
...
+
+  * course_id => String field, course identifier to allow course rating.
+
+3. Create ``RateSection`` model with the following fields.
+
+  * author => Foreign key to user model, the user who set their opinion.
+
+  * status => String field with choices: Liked, Unliked and not-set.
+
+  * usage_id => String field, this is course subsection identifier.
```
johanv26, line 35: why usage_id and not block_id?

docs/decisions/0001-course-experience.rst
```diff
@@ -0,0 +1,283 @@
...
+
+Decisions
+=========
+
+1. Create a course experience module.
+2. Create ``RateCourse`` model with the following fields.
+
+  * author => Foreign key to user model, the user who set their opinion.
+
+  * status => String field with choices: Liked, Unliked and not-set.
```
johanv26, line 25: here i have some interesting discussion about if it would be a string field. Mainly it could be  lot of data consumption saving in storage `liked` for each user. this could be a boolean or (1,0). Also the not-set could be Null.

thinking in that the not-liked has a very business discussion because will the teacher see the dislikes xD. for example in instagrama everything is like xd

docs/decisions/0001-course-experience.rst
```diff
@@ -0,0 +1,283 @@
...
+that allows users to report and share their opinions about the
+course content and experience during the course.
+
+Decisions
+=========
+
+1. Create a course experience module.
+2. Create ``RateCourse`` model with the following fields.
+
+  * author => Foreign key to user model, the user who set their opinion.
```
johanv26, line 23: Foreign key to the user model or only to the username field of the model user?

docs/decisions/0001-course-experience.rst
```diff
@@ -0,0 +1,283 @@
...
+
+6. Set model constrains, each model should be unique per sub-sections and user or course and user.
+7. Create API views for each model.
+
+
+API Examples for sub-sections
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+  GET request
+
+  /eox-nelp/api/experience/v1/rate/sections/<usage-id>/
```
johanv26, line 81: here the response is filtered by the user so the response would have that form. I mean the status of a specific user.

but with this form, it could return the number of likes of the usage_id(block_id) or that would be another endpoint?

docs/decisions/0001-course-experience.rst
andrey-canon: why do you think it's better block_id ?

docs/decisions/0001-course-experience.rst
andrey-canon: I think a boolean is a good option, the content that the instructor or student will see is out of the scope of this proposal

docs/decisions/0001-course-experience.rst
andrey-canon: Foreign key to user model

docs/decisions/0001-course-experience.rst
andrey-canon: first sentence

docs/decisions/0001-course-experience.rst
johanv26: because the definition is using block_id. https://github.com/eduNEXT/eox-nelp/blame/5fe2d6c2436c29a1b27ceb4ac6b856ebdb19e1e1/docs/decisions/0001-course-experience.rst#L91
Also, I remember that the is a phrase like "everything is an xblock"

andrey-canon: after a small research I think a better option is `item_id`, because the current endpoint `/api/courseware/sequence/{usage_key}` returns this field, and that specific field  is currently used, in the microfrontend, to generate the iframe, so if we are lucky we can take advantage of that in the new react components

johanv26 (approved):

PR #25, opened 2023-02-21
Jlc/quality improvement

johanv26:
# Description

Add the style, quality changes, and configuration to use  `make quality` and `make complexity`.
Passed:
- pycodestyle
- pylint
- isort
- flake 8 

**extra**: for the future may be we would consider these two small but important changes to refactor or leave the same. Now pylint skip them.
https://github.com/eduNEXT/eox-nelp/pull/25/commits/bbc9b9b51118b2db5c81da6509c5708da057f7d2
### quality
Peek 2023-02-21 11-54
### complexity
Peek 2023-02-21 11-56


setup.cfg
```diff
@@ -3,18 +3,15 @@ current_version = 1.0.0
...
 [pycodestyle]
 count = False
-ignore = E501, W503
+ignore = E501, W503, R903
 max-line-length = 120
 statistics = True
 
 [flake8]
 max-line-length = 120
+ignore = R903, F401, F403
```
andrey-canon, line 14: I think you can disable that for specific files, since the 403 error should be only for backends  
https://stackoverflow.com/questions/59167405/flake8-ignore-only-f401-rule-in-entire-file

and try to disable for the specific line in the  401 case

eox_nelp/middleware.py
```diff
@@ -131,6 +130,7 @@ def handler(form_instance, form_desc, required=True):
             """Wrapper of https://github.com/eduNEXT/edunext-platform/blob/ednx-release/
             mango.master.nelp/openedx/core/djangoapps/user_authn/views/registration_form.py#L657
             """
-            form_instance._add_field_with_configurable_select_options(field, label, form_desc, required=required)
-
+            form_instance._add_field_with_configurable_select_options(  # pylint: disable=protected-access
+                field, label, form_desc, required=required
```
andrey-canon, line 134: I prefer to split this in multiple lines, but if you don't want to do that, just add the trailing comma

setup.cfg
```diff
@@ -24,261 +21,204 @@ indent = '    '
...
-	basestring-builtin,
-	buffer-builtin,
-	cmp-builtin,
-	cmp-method,
-	coerce-builtin,
-	coerce-method,
-	delslice-method,
-	dict-iter-method,
-	dict-view-method,
+	consider-using-f-string,
```
andrey-canon, line 214: why not?

setup.cfg
```diff
@@ -24,261 +21,204 @@ indent = '    '
...
-	print-statement,
-	raising-string,
-	range-builtin-not-iterating,
-	raw_input-builtin,
-	reduce-builtin,
-	reload-builtin,
-	round-builtin,
-	setslice-method,
-	standarderror-builtin,
+	no-else-return,
```
andrey-canon, line 221: I like this, no idea why you are disabling this

setup.cfg
```diff
@@ -24,261 +21,204 @@ indent = '    '
...
+	file-ignored,
 	fixme,
-	getslice-method,
-	hex-method,
-	import-star-module-level,
-	indexing-exception,
-	input-builtin,
-	intern-builtin,
+	global-statement,
+	invalid-name,
```
andrey-canon, line 219: :(

andrey-canon (changes requested):
there are a lot of rules that you disabled ,and probably for open-edx makes sense but in this context I don't see the point, instead of removing we should try to activate as many as possible

setup.cfg
```diff
@@ -11,7 +11,11 @@ statistics = True
 
 [flake8]
 max-line-length = 120
-ignore = R903, F401, F403
+ignore = R903
+per-file-ignores =
```
andrey-canon, line 15: I thought the issue was related with the backends but the files are others in this case I prefer disable that per line as I mentioned in the previous comment for 401

setup.cfg
```diff
@@ -3,18 +3,18 @@ current_version = 1.0.0
...
+ignore = E501, W503, R903
 max-line-length = 120
 statistics = True
 
 [flake8]
 max-line-length = 120
+ignore = R903
+per-file-ignores =
+    */__init__.py: F403
+	*/settings/*.py: F403
```
andrey-canon, line 17: I think you forgot to remove this

eox_nelp/middleware.py
johanv26: [01a2ea4](https://github.com/eduNEXT/eox-nelp/pull/25/commits/01a2ea46a3176b2a6ccf6a7af3dff33ab9fb828c)

setup.cfg
johanv26: i prefered to keep the pylint default rules, sure seems is better to style or recommend for everything. After this commit [3f252b8](https://github.com/eduNEXT/eox-nelp/pull/25/commits/3f252b85d43928eaf42c235ce6ca27704dd978c6)I started styling the last steps.
Peek 2023-02-21 15-13

johanv26:
@andrey-canon I have passed the pylint with the new rules, the only thing is a doubt in this change:
https://github.com/eduNEXT/eox-nelp/pull/25/commits/5a0075cf87842afcd2cb7c2c3645495129b852b2
All the project use disable pylint string line by control-message and not error-code. 
the problem here is that I'm not sure to pass all the control messages because it exceeds 120 chars.
image

the commit is a solution but error codes are not too explicit. Another solution is to add all that 3 import error codes at the top of the file but will apply to all the apps.py file.

What do you think?

setup.cfg
johanv26: 403 is related to import `*` and that causes problem in flake8.
image

andrey-canon:
> @andrey-canon I have passed the pylint with the new rules, the only thing is a doubt in this change: [5a0075c](https://github.com/eduNEXT/eox-nelp/commit/5a0075cf87842afcd2cb7c2c3645495129b852b2) All the project use disable pylint string line by control-message and not error-code. the problem here is that I'm not sure to pass all the control messages because it exceeds 120 chars. image
> 
> the commit is a solution but error codes are not too explicit. Another solution is to add all that 3 import error codes at the top of the file but will apply to all the apps.py file.
> 
> What do you think?

I think you can add after the docstring, in that way that just apply to the method instead of the whole file

johanv26:
> I think you can add after the docstring, in that way that just apply to the method instead of the whole file

thanks!
[9697ef7](https://github.com/eduNEXT/eox-nelp/pull/25/commits/9697ef79483c9addfc97d0d60b4d386902869d63)

At this point again the commands are passing.

setup.cfg
johanv26: [68411cb](https://github.com/eduNEXT/eox-nelp/pull/25/commits/68411cbcd9a3ba36e97bf3bef34ebf4fb3726449)

andrey-canon (approved):
I think this fixes the most of the quality issues, so if something is missing we can fix that in the future

PR #26, opened 2023-02-22
Jlc/test init

johanv26:
# Description
let the test section run for the repo.
- Configurated and cleaned `common.py` and `test.py` settings file.
- Added test backend for each backend of eox-nelp.
- Added mako requirement due eox-theming package. Added eox-theming backend config vars  due they could be imported via plugin_settings method. https://github.com/eduNEXT/eox-theming/blob/master/eox_theming/settings/test.py#L45
- Add first mini test for repo `eox_nelp/tests/test_principal_views.py`

## Functionality
Coverage is working around 69 and 70.
Peek 2023-02-22 15-31


johanv26:
@andrey-canon take a look, and also what do you thing of 
[0740732](https://github.com/eduNEXT/eox-nelp/pull/26/commits/0740732f2956dbe829bba1ab9764a598751da89f)
This increase in 1 the coverage.

eox_nelp/settings/test.py
```diff
@@ -27,48 +27,40 @@ def plugin_settings(settings):  # pylint: disable=function-redefined
...
-    'DEFAULT': 'software_secure',
-}
+SILENCED_SYSTEM_CHECKS = [
+    'admin.E108', 'admin.E403', 'admin.E406', 'admin.E408', 'admin.E409', 'admin.E410', 'admin.E002', 'admin.E035',
+]
+
+
+# ------------external plugins configuration backends----------------
+
+EOX_THEMING_CONFIGURATION_HELPER_BACKEND = 'eox_theming.edxapp_wrapper.backends.j_configuration_helpers_tests'
```
andrey-canon, line 54: it is not possible to call eox-theming  plugin_setting method ?

eox_nelp/settings/test.py
```diff
@@ -27,48 +27,40 @@ def plugin_settings(settings):  # pylint: disable=function-redefined
...
+EOX_THEMING_THEMING_HELPER_BACKEND = 'eox_theming.edxapp_wrapper.backends.j_theming_helpers_tests'
+EOX_THEMING_EDXMAKO_BACKEND = 'eox_theming.edxapp_wrapper.backends.j_mako_tests'
+EOX_THEMING_BASE_LOADER_BACKEND = 'eox_theming.edxapp_wrapper.backends.j_loaders_tests'
+EOX_THEMING_CONFIG_SOURCES = [
```
andrey-canon, line 58: for sure for this case we only need `from_django_settings`

eox_nelp/edxapp_wrapper/test_backends/site_configuration_m_v1.py
```diff
@@ -0,0 +1,10 @@
+"""Backend test abstraction."""
+from eox_nelp.edxapp_wrapper.test_backends import create_test_model
+
+
+def get_configuration_helpers():
+    """Return test class.
+    Returns:
+        Mock class.
+    """
+    return create_test_model('configuration_helpers', 'eox_nelp', __package__)
```
andrey-canon, line 10: configuration helpers is not a model,  so I'm not sure if this line will works in the future

eox_nelp/settings/common.py
```diff
@@ -3,15 +3,15 @@
...
 SECRET_KEY = 'a-not-to-be-trusted-secret-key'
-INSTALLED_APPS = (
+INSTALLED_APPS = [
+    'django.contrib.admin',
     'django.contrib.auth',
     'django.contrib.contenttypes',
     'django.contrib.sessions',
-)
-EOX_AUDIT_MODEL_APP = 'eox_audit_model.apps.EoxAuditModelConfig'
+    'eox_nelp',
```
andrey-canon, line 12: don't we need `eox-theming` ?

eox_nelp/tests/test_principal_views.py
```diff
@@ -0,0 +1,27 @@
...
+Test views file.
+"""
+
+from django.test import TestCase
+from django.urls import reverse
+
+import eox_nelp
+
+
+class EOXInfoTestCase(TestCase):
```
andrey-canon, line 11: my concern is do we need two different tests here ? I mean there is a reason  to create two different tests when we can test both cases in just one method and the result  is always the same , or the version won't appear if `ALLOWED_HOST` is the default value

eox_nelp/tests/test_principal_views.py
```diff
@@ -0,0 +1,27 @@
...
+    """
+    Test for eox-info view.
+    """
+
+    def test_view_info_accesible(self):
+        """
+        Should get a successful answer
+        """
```
andrey-canon, line 19: I like to add expected behavior section
```
        Expected behavior:
            - Return expected content.
            - Status code 200.
        """
```

eox_nelp/settings/test.py
johanv26: Here I tried  https://github.com/eduNEXT/eox-nelp/pull/26/commits/493ed7d8975d598aac0b3a2b1113e0670a813065#diff-82e1a50636c1dd06357ba971d1ff85e6a6f09184b02208e3c3933b5aec9fce9bR51
But there is a problem related to the backends in the test settings being defined outside the `plugin_setting` method. eg eox-theming: https://github.com/eduNEXT/eox-theming/blob/master/eox_theming/settings/test.py#L45 . So the test also asked me again for the backends. 
For that reason, if you want to import the backends from  plugin setting from `test.py` file, the package should be defined inside the scope of `plugin_settings` function https://github.com/openedx/edx-django-utils/blob/master/edx_django_utils/plugins/plugin_settings.py#L20. 
You can see here eox-core do it https://github.com/eduNEXT/eox-core/blob/master/eox_core/settings/test.py#L14

eox_nelp/settings/common.py
johanv26: I set it in the past, but I  test without it, and works. Mainly it would be that mfe_config_view is not tested complete.
image
Furthermore, the use of eox-theming is not like a django app is more like a library like bs4 because we only use a method from it.
https://github.com/eduNEXT/eox-nelp/blob/master/eox_nelp/mfe_config_api/v1/views.py#L7
should we add it?
also eox-tenant is requirement but we don't add it as installed app.

eox_nelp/tests/test_principal_views.py
johanv26: yes but maybe could be a case where eox-info URL is working respond 200 but in a different version.
or is improbably? xD

eox_nelp/settings/common.py
andrey-canon: no, we shouldn't

eox_nelp/tests/test_principal_views.py
andrey-canon: I think that view has not changed in the last  3 years in another plugins

eox_nelp/tests/test_principal_views.py
johanv26: https://github.com/eduNEXT/eox-nelp/pull/26/commits/3defcd0ab909bac5fe867e046b75a86a4472bf7c

eox_nelp/settings/test.py
johanv26: [3358e93](https://github.com/eduNEXT/eox-nelp/pull/26/commits/3358e93202a20426a18e2bb352b0697f0fc54585)

eox_nelp/tests/test_principal_views.py
andrey-canon: yes, I prefer that test case, just one last comment, the name should be just `test_views.py` generally this follow the same structure of the container folder.

eox_nelp/edxapp_wrapper/test_backends/site_configuration_m_v1.py
johanv26: [7adf67d](https://github.com/eduNEXT/eox-nelp/pull/26/commits/7adf67d823de4083a57dc900b5855c179c91aaa2)

johanv26:
> @andrey-canon take a look, and also what do you thing of [0740732](https://github.com/eduNEXT/eox-nelp/pull/26/commits/0740732f2956dbe829bba1ab9764a598751da89f) This increase in 1 the coverage.

I have updated with recommendations, this moment I only have this doubt.

eox_nelp/tests/test_principal_views.py
johanv26: yes but, this are only views of the first level folder of the plugin. Dont test all views of the plugin. Is ok to use `test_views.py`?

andrey-canon:
> > @andrey-canon take a look, and also what do you thing of [0740732](https://github.com/eduNEXT/eox-nelp/pull/26/commits/0740732f2956dbe829bba1ab9764a598751da89f) This increase in 1 the coverage.
> 
> I have updated with recommendations, this moment I only have this doubt.

omit those files for me is ok

eox_nelp/tests/test_principal_views.py
andrey-canon: thast file should just test the views at the first level, if there are more views files those could be tested inside the parent folder or you can add a folder with the same name inside the test folder.

so if you want to test course_api/views, you can create a test folder inside course_api, and then create another test_views.py file 

or you can create a course_api inside current test folder and the the test_views file

johanv26: knowing that test_views will be referred to views of each module or folder, I agree.
[201584b](https://github.com/eduNEXT/eox-nelp/pull/26/commits/201584b4f0848f3bf53a3926ba4d180c5aedba6f)

andrey-canon (approved):

PR #28, opened 2023-02-27
feat: add course experience model based on 0001-course-experience ADR

andrey-canon:
## Description
Add course experience model based on 0001-course-experience ADR


eox_nelp/course_experience/models.py
```diff
@@ -0,0 +1,114 @@
...
+    fields:
+        author<Foreignkey>: Makes reference to the user record associated with the reason.
+        reason<CharField>: Store report reason as a code, e.g SC => Sexual Content.
+    """
+    REPORT_REASONS = [
+        ("SC", "Sexual content"),
+        ("GV", "Graphic violence"),
+        ("HA", "Hateful or abusive content"),
+        ("CI", "Copycat or impersonation"),
+        ("OO", "Other objection"),
```
johanv26, line 44: I thought another objection would be a variable or something the user could change or give more information about.
by the way, at the moment we can keep it this way.

johanv26 (commented):
here maybe is not the moment but we could be at a time before filling the models in db. 
In the section models, I found an interesting relation to the courseware modulstore  admin.
Peek 2023-02-27 15-20
As you see here they have two extra models for the section or "modules":  `module_type` and `course_id`. This is something we can use as a reference or compare.  
eg, adding the course_id field helps to filter, but also you are storing repeated course_id that could be extracted in a different way from the module key(item_id).
Is only something to consider, because the current pr seems minimalist something that also is good.
Also for example the admin doesn't have a search bar to maybe count all section likes related to a specific course.

andrey-canon:
> other difference for example is that the admin doesn'



> here maybe is not the moment but we could be at a time before filling the models in db. In the section models, I found an interesting relation to the courseware modulstore admin. Peek 2023-02-27 15-20 [ Peek 2023-02-27 15-20 ](https://user-images.githubusercontent.com/51926076/221677587-7c0bfa0a-51c7-40ae-b952-ea8e2c063b1e.gif) [ ](https://user-images.githubusercontent.com/51926076/221677587-7c0bfa0a-51c7-40ae-b952-ea8e2c063b1e.gif) As you see here they have two extra models for the section or "modules": `module_type` and `course_id`. This is something we can use as a reference or compare. eg, adding the course_id field helps to filter, but also you are storing repeated course_id that could be extracted in a different way from the module key(item_id). another difference for example is that the admin doesn't have a search bar to maybe count all section likes related to a specific course. Is only something to consider, because the current pr seems minimalist something that also is good.

@johanv26  What do you mean ? do you want to change course_id to a foreignkey of  courseoverview and item_id  to ...?

johanv26:
No I am only showing the relation with the other model and principally ask for for  your opinion about:
- if we should add that field of course_id in the section models.
- change item_id by module key.
- is important to keep or use module type in the section models. I think not xD but is better to ask.

johanv26:
@andrey-canon Itestes and seem the course_id is matching ok, will the sections be renamed to units?

andrey-canon:
> @andrey-canon I tested and seem the course_id is matching ok, will the sections be renamed to units?

yes, I forgot that part, thanks for the reminder

johanv26 (approved):

PR #43, opened 2023-04-17
feat: add notifications models and its admin file

andrey-canon:
## Description

This address this [issue](https://edunext.atlassian.net/browse/FUTUREX-394)

## Testing instructions
1. run ./manage.py lms migrate eox_nelp
2. Go to admin `admin/eox_nelp/upcomingcourseduedate/`
3. Create multiple records with different values
4. Test search funtionality
5. Test filters
6. Select some records a select the action notify selected notifications
7. Check logs severs

## Checklist for Merge

- [ ] Updated documentation
- [ ] Rebased master/main
- [ ] Squashed commits


eox_nelp/notifications/admin.py
```diff
@@ -0,0 +1,98 @@
...
+                    self.lookup_kwarg_since: today,
+                    self.lookup_kwarg_until: today + datetime.timedelta(days=7),
+                },
+            ),
+            (
+                _("Next month"),
+                {
+                    self.lookup_kwarg_since: today,
+                    self.lookup_kwarg_until: today + datetime.timedelta(days=30),
+                },
```
johanv26, line 61: What do you think of adding more than 30 days filter or its innecessary?

andrey-canon: Could you give me an example, where that could be necessary ?

eox_nelp/notifications/admin.py
```diff
@@ -0,0 +1,98 @@
...
+
+    def __init__(self, field, request, params, model, model_admin, field_path):  # pylint: disable=too-many-arguments
+        """Custom init method to modify inherited options"""
+        super().__init__(field, request, params, model, model_admin, field_path)
+        today = timezone.now()
+
+        self.links = (
+            (_("Any date"), {}),
+            (
+                _("Next day"),
```
johanv26, line 43: This translations already exists?

eox_nelp/notifications/admin.py
johanv26: No looking more the filters are not Mutual exclusivity so exists intersection between the next day and the next week for example.
So for that case, I was thinking of filtering by more than one month but not including between today and the month.  But seems to be not the purpose of the filter, so forget it.

johanv26 (commented):
I approve. The only thing for future work is to check some ar translations, but for the moment is working nicely.

johanv26 (approved):

eox_nelp/notifications/models.py
```diff
@@ -0,0 +1,45 @@
...
+
+    Fields:
+        course_id<Foreignkey>: Reference to a specific course.
+        location_id<CharField>: Store subsection id.
+        due_date<DateTimeField>: Subsection due date.
+        notification_date<DateTimeField>: Date when the notification must be sent.
+        sent<BooleanField>: Indicates if the notification was already sent.
+    """
+    course_id = models.ForeignKey(CourseOverview, null=False, on_delete=models.CASCADE)
+    location_id = models.CharField(max_length=255, null=False)
```
johanv26, line 28: Do you decide to avoid using the usage_key model field? I thinks because is becomes easier to develop and avoid crazy 500 management of usakge key shape?

andrey-canon: Indeed I thought the location that comes from the subsection was a simple string, but now I'm working with the receiver  and the location is an usage id therefore I'm going to update the model :(

andrey-canon:
> I approve. The only thing for future work is to check some ar translations, but for the moment is working nicely. 

@johanv26  please delete those records before running new migrations to avoid 500 errors

eox_nelp/notifications/models.py
johanv26: checking this also I think you can change course_id to course?
this due is strange access in this way.
image

johanv26: course enrollments do it in that way.
image

johanv26 (commented):
Before squash it would be only the 002 migration for upcomigcourseduedate, or would be until the number 004?

andrey-canon:
> Before squash it would be only the 002 migration for upcomigcourseduedate, or would be until the number 004?

just 002

PR #44, opened 2023-04-18
Jlc/notify course due dates

johanv26:
## Description

This manages the logic for notification task and also add the command to check and notify upcoming course due dates.
- added backend for course-email-template
- you need to configure a template in bulk-email admin with name: `base-notification-template`
image

### After
Peek 2023-04-20 13-01
image

## Checklist for Merge

- [x] Tested in a remote environment
- [ ] Updated documentation
- [ ] Rebased master/main
- [ ] Squashed commits

### DIFF
--- a/eox_nelp/notifications/notify_course_due_date.py
+++ b/eox_nelp/notifications/notify_course_due_date.py
@@ -1,0 +1,107 @@
+"""Notify course date module. This manage the delivery of an
+upcoming_course_due_date
+"""
+import logging
+
+from django.utils import timezone
+
+from eox_nelp.edxapp_wrapper.bulk_email import CourseEmailTemplate, get_course_email_context
+from eox_nelp.edxapp_wrapper.student import CourseEnrollment
+from eox_nelp.notifications.utils import (
+    get_xblock_course_from_usage_key_string,
+    get_xblock_from_usage_key_string,
+    send_email_multialternative,
+)
+
+logger = logging.getLogger(__name__)
+
+
+def notify_upcoming_course_due_date(upcoming_course_due_date):
+    """Send the correspongind email of an UpcomingCourseDate object.
+    This is using the bulk-email template that already exists.
+    Args:
+        upcoming_course_due_date (UpcomingCourseDate): UpcomingCourseDate object to sent emails.
+    """
+    subsection_key_string = str(upcoming_course_due_date.location_id)
+    enrollments_qs = CourseEnrollment.objects.filter(course__id=upcoming_course_due_date.course_id, is_active=True)
+    recipient_emails = [enrollment.user.email for enrollment in enrollments_qs if enrollment.user.email]
+
+    subsection_xblock = get_xblock_from_usage_key_string(subsection_key_string)
+    course_xblock = get_xblock_course_from_usage_key_string(subsection_key_string)
+
+    if not (subsection_xblock and recipient_emails and course_xblock):
+        logger.info(
+            "is missing one of the following: subsection-xblock(%s),recipient_emails(%s),course_xblock(%s)",
+            subsection_xblock, recipient_emails, course_xblock
+        )
+        return
+
+    rendered_email_msgs = generate_email_messages(subsection_xblock, course_xblock)
+    # to do, add tenant aware settings for context and particular course notification template.
+
+    sent = send_email_multialternative(**rendered_email_msgs, recipient_emails=recipient_emails)
+
+    if sent:
+        upcoming_course_due_date.sent = True
+        upcoming_course_due_date.save()
+        logger.info(
+            "%s emails sent related the upcoming_course_due_date: %s -------",
+            sent,
+            upcoming_course_due_date.__dict__,
+        )
+    else:
+        logger.info(
+            "------- No emails were sent!!!, the email function returned %s -------",
+            sent,
+        )
+
+
+def generate_email_messages(subsection_xblock, course_xblock):
+    """Generate the messages of the email with the data of the subsection and
+    course xblock. Return  a dict with the the keys subject, plaintext_msg and html_msg
+    of the email to sent.
+
+        Args:
+            subsection_xblock (subsectionXblock)
+            course_xblock (courseXblock)
+
+        Returns:
+            (dict): Dict with the messages rendered of the email.
+            It contains: subject, plaintext_msg and html_msg
+    """
+    email_context = get_course_email_context(course_xblock)
+    course_email_template = CourseEmailTemplate.get_template(name='base-notification-template')
+    subsection_name = subsection_xblock.display_name
+    course_name = course_xblock.display_name
+    subsection_due_date = str(subsection_xblock.due)
+    delta = subsection_xblock.due - timezone.now()
+    time_remaining = (
+        f"{delta.days} days, {delta.seconds // 3600:02d} hours, "
+        f"and {(delta.seconds // 60) % 60:02d} mins"
+    )
+    text_message = (
+        f"The course {course_name} has a notification. "
+        f"The subsection {subsection_name} is coming soon. "
+        f"The due date of this subsection is UTC {subsection_due_date}. "
+        f"There are approx {time_remaining} for the due date of the subsection."
+    )
+
+    html_message = (
+        f"<p>The course {course_name} has a notification. "
+        f"The subsection {subsection_name} is coming soon.</p>\n"
+        f"<p>The due date of this subsection is UTC {subsection_due_date}. "
+        f"There are approx {time_remaining} for the due date of the subsection.</p>"
+    )
+
+    subject = (
+        f"[Futurex platform] Notification due date(left {delta.days} days) of subsection {subsection_name} "
+        f"from course {course_name}"
+    )
+    plaintext_msg = course_email_template.render_plaintext(text_message, email_context)
+    html_msg = course_email_template.render_htmltext(html_message, email_context)
+
+    return {
+        "subject": subject,
+        "plaintext_msg": plaintext_msg,
+        "html_msg": html_msg,
+    }
--- a/eox_nelp/notifications/tests/test_models.py
+++ b/eox_nelp/notifications/tests/test_models.py
@@ -6,6 +6,7 @@ Classes:
 import unittest
 
 from django.utils import timezone
+from mock import patch
 
 from eox_nelp.edxapp_wrapper.course_overviews import CourseOverview
 from eox_nelp.notifications import models
@@ -14,12 +15,15 @@ from eox_nelp.notifications.models import UpcomingCourseDueDate
 
 class UpcomingCourseDueDateTestCase(unittest.TestCase):
     """Test class for model UpcomingCourseDueDate"""
-
-    def test_nofity_log(self):
-        """Test that the notify model logs progress message.
+    @patch("eox_nelp.notifications.tasks.notify_upcoming_course_due_date_by_id")
+    def test_send_notifications(self, notify_upcoming_course_due_date_by_id_mock):
+        """Test that the notify method of UpcomingCourseDueDate
+        call the task `notify_upcoming_course_due_date_by_id` and the
+        correct logs progress message.
 
         Expected behavior:
             - Log progress message successfully.
+            - notify_upcoming_course_due_date_by_id method was called.
         """
         instance, _ = UpcomingCourseDueDate.objects.get_or_create(  # pylint: disable=no-member
             course=CourseOverview.objects.create(id="course-v1:test+Cx105+2022_T4"),
@@ -33,3 +37,4 @@ class UpcomingCourseDueDateTestCase(unittest.TestCase):
             instance.notify()
 
         self.assertEqual(logs.output, [f"INFO:{models.__name__}:{processing_log}"])
+        notify_upcoming_course_due_date_by_id_mock.delay.assert_called_with(instance.id)
--- a/eox_nelp/notifications/tests/test_notify_course_due_date.py
+++ b/eox_nelp/notifications/tests/test_notify_course_due_date.py
@@ -1,0 +1,216 @@
+"""This file contains all the test for notifications/notify_course_due_date.py file.
+
+Classes:
+    NotifyUpcomingCourseDueDateTestCase: Test `notify_upcoming_course_due_date` method.
+"""
+from datetime import timedelta
+
+from ddt import data, ddt
+from django.test import TestCase
+from django.utils import timezone
+from mock import patch
+
+from eox_nelp.edxapp_wrapper.course_overviews import CourseOverview
+from eox_nelp.notifications import notify_course_due_date
+from eox_nelp.notifications.models import UpcomingCourseDueDate
+from eox_nelp.notifications.notify_course_due_date import notify_upcoming_course_due_date
+from eox_nelp.tests.utils import generate_list_mock_data
+
+FALSY_SENT_MAILS_VALUES = [0, "", None, [], False, {}, ()]
+
+
+@ddt
+class NotifyUpcomingCourseDueDateTestCase(TestCase):
+    """Test class for function `notify_upcoming_course_due_date`."""
+
+    def setUp(self):
+        """Set common conditions for test cases."""
+        self.patchers = [
+            patch("eox_nelp.notifications.notify_course_due_date.send_email_multialternative"),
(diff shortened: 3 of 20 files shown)


eox_nelp/notifications/notify_course_due_date.py
```diff
@@ -0,0 +1,78 @@
...
+    """
+    #  pylint: disable=too-many-locals
+    subsection_key_string = str(upcoming_course_due_date.location_id)
+    enrollments_qs = CourseEnrollment.objects.filter(course__id=upcoming_course_due_date.course_id, is_active=True)
+    recipient_emails = [enrollment.user.email for enrollment in enrollments_qs if enrollment.user.email]
+
+    subsection_xblock = get_xblock_from_usage_key_string(subsection_key_string)
+    course_xblock = get_xblock_course_from_usage_key_string(subsection_key_string)
+
+    if (not subsection_xblock) or (not recipient_emails) or (not course_xblock):
```
andrey-canon, line 33: I know I said use `or` but  you mentioned something with `and`  and now I'm thinking in something like this suggestion
```suggestion
    if not (subsection_xblock and recipient_emails and course_xblock):
```

eox_nelp/notifications/tests/test_models.py
```diff
@@ -14,8 +15,8 @@
 
 class UpcomingCourseDueDateTestCase(unittest.TestCase):
     """Test class for model UpcomingCourseDueDate"""
-
-    def test_nofity_log(self):
+    @patch("eox_nelp.notifications.tasks.notify_upcoming_course_due_date_by_id")
+    def test_nofity_log(self, *args):  # pylint: disable=unused-argument
```
andrey-canon, line 19: I think we should improve this test and rename it to test `test_send_notifications` and verify that the task was called

eox_nelp/notifications/tests/test_tasks.py
```diff
@@ -374,3 +375,23 @@ def _generate_subsections(self, subsections_data):
             subsections.append(subsection)
 
         return subsections
+
+
+class NotifyUpcomingCourseDueDateByIdTestCase(unittest.TestCase):
```
andrey-canon, line 380: the task could raise `DoesNotExist` exception, should we handle that ?

eox_nelp/notifications/tests/test_notify_course_due_date.py
```diff
@@ -0,0 +1,138 @@
...
+            - email_key_words are presented in the `text_message` arg (0) of 2nd call of course_email_template.
+            - email_key_words are presented in the `html_message` arg (0) of 3rd call of course_email_template.
+            - course_name is presented in `subject` arg(0) of the call of send_email_multialternative.
+            - subsection_name is presented in `subject` arg(0) of the call of send_email_multialternative.
+            - recipient list is presented in `recipient_emails` arg(3) of the call of send_email_multialternative.
+            - The email was sent from the correct email address (if specified)
+            - sent attribute of instance of `UpcomingCourseDueDate` after the process is True.
+            - Log successful of sent message.
+        """
+        UpcomingCourseDueDate.objects.all().delete()  # pylint: disable=no-member
```
andrey-canon, line 66: why do you need this?

eox_nelp/tests/utils.py
```diff
@@ -0,0 +1,48 @@
+""" Module with methods to reuse in test cases"""
+from mock import Mock
+
+
+def generate_list_mock_data(data):
```
andrey-canon, line 5: if you need something like this you should consider to improve the method that already exists instead of create a new one

eox_nelp/notifications/tests/test_notify_course_due_date.py
```diff
@@ -0,0 +1,138 @@
...
+            {
+                "display_name": self.course_name,
+            },
+        ]
+
+    @patch("eox_nelp.notifications.notify_course_due_date.send_email_multialternative")
+    @patch("eox_nelp.notifications.notify_course_due_date.CourseEmailTemplate")
+    @patch("eox_nelp.notifications.notify_course_due_date.get_xblock_course_from_usage_key_string")
+    @patch("eox_nelp.notifications.notify_course_due_date.get_xblock_from_usage_key_string")
+    @patch("eox_nelp.notifications.notify_course_due_date.CourseEnrollment")
```
andrey-canon, line 45: could you consider to move this to the setUp method, check this example https://github.com/eduNEXT/eox-nelp/blob/master/eox_nelp/signals/tests/test_tasks.py#L203

eox_nelp/notifications/tests/test_notify_course_due_date.py
```diff
@@ -0,0 +1,138 @@
...
+            self.course_name,
+            self.subsection_name,
+            str(self.subsection_due_date),
+        ]
+
+        with self.assertLogs(notify_course_due_date.__name__, level="INFO") as logs:
+            notify_upcoming_course_due_date(instance)
+
+        for word in email_key_words:
+            self.assertIn(word, course_email_template_mock.mock_calls[1][1][0])
```
andrey-canon, line 88: why are you testing in this way that the word is in `text_message` when you can use an `assert_called_with ` and test the whole message instead of just one word ?

eox_nelp/notifications/tests/test_notify_course_due_date.py
```diff
@@ -0,0 +1,138 @@
...
+            self.assertIn(word, course_email_template_mock.mock_calls[2][1][0])
+        self.assertIn(self.course_name, send_email_multialternative_mock.mock_calls[0][1][0])
+        self.assertIn(self.subsection_name, send_email_multialternative_mock.mock_calls[0][1][0])
+        self.assertListEqual(self.recipients_list, send_email_multialternative_mock.mock_calls[0][1][3])
+        self.assertTrue(instance.sent)
```
andrey-canon, line 93: you are not testing when `sent` is false

andrey-canon (changes requested):

eox_nelp/notifications/notify_course_due_date.py
```diff
@@ -0,0 +1,78 @@
...
+logger = logging.getLogger(__name__)
+
+
+def notify_upcoming_course_due_date(upcoming_course_due_date):
+    """Send the correspongind email of an UpcomingCourseDate object.
+    This is using the bulk-email template that already exists.
+    Args:
+        upcoming_course_due_date (UpcomingCourseDate): UpcomingCourseDate object to sent emails.
+    """
+    #  pylint: disable=too-many-locals
```
andrey-canon, line 25: if pylint raises this warning you should consider to stop using so many variables or split the method

eox_nelp/notifications/tests/test_notify_course_due_date.py
johanv26: Here ?
https://github.com/eduNEXT/eox-nelp/pull/44/files/fda7e2bc2e4ffc5055707b38b6c56845a3f515c2#diff-f2c383ced14e27737fb812392f5da8f4d4811351bc15cad4698a7052ac606953R134

eox_nelp/notifications/tests/test_notify_course_due_date.py
johanv26: I cant use `assert_called_with` because it use the last call args.  Maybe I can change with the  `assert_any_call` https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.assert_any_call.
No wy the problem is that I cant generate all the message because there is a time now generated, in term of seconds would change.  https://github.com/eduNEXT/eox-nelp/blob/fda7e2bc2e4ffc5055707b38b6c56845a3f515c2/eox_nelp/notifications/notify_course_due_date.py#L48

johanv26: https://github.com/eduNEXT/eox-nelp/blob/fda7e2bc2e4ffc5055707b38b6c56845a3f515c2/eox_nelp/notifications/notify_course_due_date.py#L56

eox_nelp/notifications/tests/test_notify_course_due_date.py
johanv26: Because I am creating the same instance of UpcomingCourseDueDate, with the same data. In the first one I pass the sent to true, in the second verify the False. So it could change depending the order or running functions. To avoid that, I ensure remove all. Othe option is to create and instance with different data.

eox_nelp/notifications/utils.py
```diff
@@ -0,0 +1,58 @@
...
+
+    Returns:
+        xblock_course: xblock of courses object. If not found return None.
+    """
+    try:
+        usage_key = UsageKey.from_string(usage_key_string)
+        store = modulestore()
+        xblock_course = store.get_course(usage_key.course_key, depth=None)
+        return xblock_course
+    except InvalidKeyError:
```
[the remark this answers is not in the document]
johanv26: No, if there is not course  that method returns None

eox_nelp/tests/utils.py
johanv26: Yes, but I don't want it to call it from https://github.com/eduNEXT/eox-nelp/blob/master/eox_nelp/notifications/tests/test_tasks.py because the method could be extended to other modules.
And as you see it I don't use it to `_generate subsection`, I use it to generate whatever list mock like course enrollments.

eox_nelp/notifications/tests/test_notify_course_due_date.py
andrey-canon: `get_or_create` or `update_or_create ` because that makes me think that your test only works with an empty database

eox_nelp/notifications/tests/test_notify_course_due_date.py
andrey-canon: so your method is calling this multiple times ?

eox_nelp/notifications/tests/test_notify_course_due_date.py
andrey-canon: that tests when `get_xblock_course_from_usage_key_string` returns None, I refer when sent = send_email_multialternative(subject, plaintext_msg, html_msg, recipient_emails) returns None or False

eox_nelp/tests/utils.py
andrey-canon: same comment

eox_nelp/notifications/tests/test_notify_course_due_date.py
johanv26: ok let me added a test to add hehe

eox_nelp/notifications/tests/test_notify_course_due_date.py
johanv26: yes, the course_email_template_mock yes. But the specific method not. Let me check if I could tried with that.
https://github.com/eduNEXT/eox-nelp/blob/fda7e2bc2e4ffc5055707b38b6c56845a3f515c2/eox_nelp/notifications/notify_course_due_date.py#L67-L68

eox_nelp/tests/utils.py
johanv26: I dont understand the same comment, one thing I can do is delete _generate_subsection and call in that file generate_list_mock_data. But that is not obtrusive?

andrey-canon: I mean this method does the same thing that an older method does but in a better and wide way, so you should consider to delete the old one and improve the general code not just your implementation, that help us to mantain less code

eox_nelp/notifications/tests/test_notify_course_due_date.py
johanv26: But the problem is the same, I cant replicate `text_message` or `html_message` due delta calculated when is called  the function.

eox_nelp/notifications/notify_course_due_date.py
```diff
@@ -0,0 +1,78 @@
...
+    subsection_key_string = str(upcoming_course_due_date.location_id)
```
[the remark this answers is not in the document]
johanv26: I like to have the string, also location_id in the first step migrations was string so I had the logic. Then change it to usage_key so for fast compatibility change what I thought was to set it to a string before the methods.
 Is Ok to have it that way, or is better to update the methods xD

eox_nelp/notifications/tests/test_notify_course_due_date.py
andrey-canon: so there are something wrong with your implementation base on your code `CourseEmailTemplate` is never called, the attribute `get_template` is called, probably that returns  a mock that  has  other mock attributes like `render_plaintext`

Your time issue don't make you wonder why are you sending time with milliseconds ? in this context is that worthy?

johanv26: it's worthy it If  I build a string with that. So assert_called_with will fail.
**https://github.com/eduNEXT/eox-nelp/blob/fda7e2bc2e4ffc5055707b38b6c56845a3f515c2/eox_nelp/notifications/notify_course_due_date.py#L60**

johanv26: Is would be different. 
image

eox_nelp/notifications/tests/test_notify_course_due_date.py
johanv26: [19b22cb](https://github.com/eduNEXT/eox-nelp/pull/44/commits/19b22cb8d382753eebe58111a49bb76ed96a3b69)

eox_nelp/notifications/tests/test_notify_course_due_date.py
johanv26: [19b22cb](https://github.com/eduNEXT/eox-nelp/pull/44/commits/19b22cb8d382753eebe58111a49bb76ed96a3b69)

eox_nelp/tests/utils.py
johanv26: [fa5e619](https://github.com/eduNEXT/eox-nelp/pull/44/commits/fa5e6198633321a62221fb2968560cd61553f3d8)

eox_nelp/notifications/tests/test_notify_course_due_date.py
johanv26: [cdd2136](https://github.com/eduNEXT/eox-nelp/pull/44/commits/cdd2136830dd5425dd813288ac834c2add3190c0)

eox_nelp/notifications/notify_course_due_date.py
johanv26: [c782c5a](https://github.com/eduNEXT/eox-nelp/pull/44/commits/c782c5a6be0ba58eb4ff465f8c256db32441b5a4)

eox_nelp/notifications/tests/test_notify_course_due_date.py
andrey-canon: same comment, in this context is that worthy? do you think the user will find that information useful? why do you think  the value  194313 is important in this context ?

eox_nelp/notifications/notify_course_due_date.py
```diff
@@ -0,0 +1,112 @@
...
+
+    rendered_email_msgs = generate_email_messages(subsection_xblock, course_xblock)
+    # to do, add tenant aware settings for context and particular course notification template.
+
+    sent = send_email_multialternative(
+        rendered_email_msgs['subject'],
+        rendered_email_msgs['plaintext_msg'],
+        rendered_email_msgs['html_msg'],
+        recipient_emails,
+    )
```
andrey-canon, line 47:
```suggestion
    sent = send_email_multialternative(
        **rendered_email_msgs,
        recipient_emails,
    )
```

andrey-canon (approved):

eox_nelp/notifications/notify_course_due_date.py
johanv26: [b6fd7ee](https://github.com/eduNEXT/eox-nelp/pull/44/commits/b6fd7ee2b22cac8fd03be0236e0196ec54eef568)

eox_nelp/notifications/tests/test_tasks.py
johanv26: https://github.com/eduNEXT/eox-nelp/pull/44/commits/4299f3197f74f8f14c154d002844a00beb73e4f9#diff-80232d77731fe3d63292ad0abfb7fece44750d6e59f2573efefc2864703b1afbR369

PR #51, opened 2023-05-31
feat: add TenantExistMiddleware

andrey-canon:
## Description

This adds a middleware that return 404 if the tenant has not been set by eox-tenant 

## Testing instructions

1. Add the TenantExist Middleware to your MIDDLEWARE setting
2. Access to a site with TenantConfig and Route, that should work normally
3. Access to a site without TenantConfig and Route, that should return 404


johanv26 (approved):
I like the compact of the middleware, and I add just a reminder not to include this middleware in studio settings because the studio does not all time has a tenant and route.
https://github.com/nelc/manifest-sagar/blob/main/plugins/common_setting_manager.yml#L401
Also, this could be in the start steps of the MIDDLEWARE list to return fast 404 xD.

PR #55, opened 2023-06-27
And/stats api view

andrey-canon:
## Description

This add a new endpoint in `/eox-nelp/api/stats/v1/tenant-stats/` that will return the basic tenant stats in a json format

```
{
    "learners": 1,
    "courses": {
        "total_courses": 4,
        "metrics": [
            {
                "id": "course-v1:patata+CS102+2023",
                "name": "PROCEDURAL SEDATION AND ANALGESIA COURSE",
                "learners": 0,
                "instructors": 1,
                "sections": 18,
                "sub_sections": 144,
                "units": 184,
                "components": {
                    "discussion": 0,
                    "drag-and-drop-v2": 0,
                    "html": 133,
                    "openassessment": 0,
                    "problem": 49,
                    "video": 0
                }
            },
            {
                "id": "course-v1:test+CS101+2022_T3",
                "name": "test",
                "learners": 1,
                "instructors": 0,
                "sections": 1,
                "sub_sections": 6,
                "units": 7,
                "components": {
                    "discussion": 0,
                    "drag-and-drop-v2": 2,
                    "html": 0,
                    "openassessment": 0,
                    "problem": 6,
                    "video": 2
                }
            },
            ...
        ]
    },
    "instructors": 1
}
```

issue https://edunext.atlassian.net/browse/FUTUREX-440

## Testing instructions
1. Go to the endpoint `/eox-nelp/api/stats/v1/tenant-stats/` and check 
2. To add components set the setting ALLOWED_VERTICAL_BLOCK_TYPES

### DIFF
--- a/eox_nelp/stats/api/v1/urls.py
+++ b/eox_nelp/stats/api/v1/urls.py
@@ -1,0 +1,15 @@
+"""
+Course API URLs
+"""
+from django.conf import settings
+from django.urls import path, re_path
+
+from eox_nelp.stats.api.v1.views import GeneralCourseStatsView, GeneralTenantStatsView
+
+app_name = "eox_nelp"  # pylint: disable=invalid-name
+
+urlpatterns = [
+    path('tenant/', GeneralTenantStatsView.as_view(), name="general-stats"),
+    path('courses/', GeneralCourseStatsView.as_view(), name="courses-stats"),
+    re_path(rf'^courses/{settings.COURSE_ID_PATTERN}', GeneralCourseStatsView.as_view(), name="course-stats"),
+]
--- a/eox_nelp/stats/api/v1/views.py
+++ b/eox_nelp/stats/api/v1/views.py
@@ -1,0 +1,145 @@
+"""Stats API v1 view file.
+
+views:
+    GeneralTenantStatsView: View that handles the general tenant stats.
+    GeneralTenantCoursesView: View that handles the general courses stats.
+"""
+from django.contrib.auth import get_user_model
+from django.http import Http404
+from rest_framework.response import Response
+from rest_framework.views import APIView
+
+from eox_nelp.stats import metrics
+
+User = get_user_model()
+
+
+class GeneralTenantStatsView(APIView):
+    """Class view. Handle general tenant stats.
+
+    ## Usage
+    The components key depends on the setting API_XBLOCK_TYPES, this should be
+    a list of strings like the following
+
+    ``` json
+    [" problem", "video", "discussion"]
+    ```
+
+    ### **GET** /eox-nelp/api/stats/v1/tenant/
+
+    **GET Response Values**
+    ``` json
+    {
+        "learners": 1,
+        "courses": 3,
+        "instructors": 2,
+        "components": {
+            "discussion": 0,
+            "drag-and-drop-v2": 0,
+            "html": 133,
+            "openassessment": 0,
+            "problem": 49,
+            "video": 0
+        }
+    }
+    ```
+    """
+
+    def get(self, request):
+        """Return general tenant stats."""
+        tenant = request.site.domain
+        courses = metrics.get_courses_metrics(tenant)
+        components = {}
+
+        for metric in courses.get("metrics", []):
+            course_components = metric.get("components", {})
+
+            for key, value in course_components.items():
+                components[key] = components.get(key, 0) + value
+
+        return Response({
+            "learners": metrics.get_learners_metric(tenant),
+            "courses": courses.get("total_courses", 0),
+            "instructors": metrics.get_instructors_metric(tenant),
+            "components": components
+        })
+
+
+class GeneralCourseStatsView(APIView):
+    """Class view that returns a list of course stats or a specific course stats.
+
+    ## Usage
+    The components key depends on the setting ALLOWED_VERTICAL_BLOCK_TYPES, this should be
+    a list of strings like the following
+
+    ``` json
+    [" problem", "video", "discussion"]
+    ```
+
+    ### **GET** /eox-nelp/api/stats/v1/courses/
+
+    **GET Response Values**
+    ``` json
+    {
+        "total_courses": 4,
+        "metrics": [
+            {
+                "id": "course-v1:patata+CS102+2023",
+                "name": "PROCEDURAL SEDATION AND ANALGESIA COURSE",
+                "learners": 0,
+                "instructors": 1,
+                "sections": 18,
+                "sub_sections": 144,
+                "units": 184,
+                "components": {
+                    "discussion": 0,
+                    "drag-and-drop-v2": 0,
+                    "html": 133,
+                    "openassessment": 0,
+                    "problem": 49,
+                    "video": 0
+                }
+            },
+            ...
+        ]
+    }
+    ```
+
+    ### **GET** /eox-nelp/api/stats/v1/courses/course-v1:potato+CS102+2023/
+
+    **GET Response Values**
+    ``` json
+    {
+        "id": "course-v1:potato+CS102+2023",
+        "name": "PROCEDURAL SEDATION AND ANALGESIA COURSE",
+        "learners": 0,
+        "instructors": 1,
+        "sections": 18,
+        "sub_sections": 144,
+        "units": 184,
+        "components": {
+            "discussion": 0,
+            "drag-and-drop-v2": 0,
+            "html": 133,
+            "openassessment": 0,
+            "problem": 49,
+            "video": 0
+        }
+    }
+    ```
+    """
+
+    def get(self, request, course_id=None):
+        """Return general course stats."""
+        tenant = request.site.domain
+
+        if course_id:
+            courses = metrics.get_cached_courses(tenant)
+            course = courses.filter(id=course_id).first()
+
+            if not course:
+                raise Http404
+
+            return Response(metrics.get_course_metrics(course.id))
+
+        return Response(metrics.get_courses_metrics(tenant))
--- a/eox_nelp/stats/decorators.py
+++ b/eox_nelp/stats/decorators.py
@@ -1,0 +1,37 @@
+"""eox-nelp Metrics file.
+
+decorators:
+    cache_method: Cache the result of the inner method.
+"""
+from django.conf import settings
+from django.core.cache import cache
+
+
+def cache_method(func):
+    """
+    Cache the function result to improve the response time.
+
+    Args:
+        func<function>: Target function to be cached.
+
+    Return:
+        <funtion>: Wrapper function.
+    """
+    def wrapper(*args, **kwargs):
+        key = f"{func.__name__}.{'-'.join(map(str, args))}.STATS_CACHE_KEY"
+        result = cache.get(key)
+
+        if result:
+            return result
+
+        result = func(*args, **kwargs)
+
+        cache.set(
+            key,
+            result,
+            timeout=getattr(settings, "STATS_SETTINGS", {}).get("STATS_TIMEOUT", 3600),
+        )
+
+        return result
+
+    return wrapper
--- a/eox_nelp/stats/metrics.py
+++ b/eox_nelp/stats/metrics.py
@@ -1,0 +1,136 @@
+"""eox-nelp Metrics file.
+
+functions:
+    get_cached_courses: Return visible courses.
+    get_course_metrics: Return the metric for the given course_key.
+    get_learners_metric: Return number of learners, for the visible courses.
+    get_instructors_metric: Return number of instructors, for the visible courses.
+    get_courses_metrics: Return metrics for the visible courses.
+"""
+from django.conf import settings
+
+from eox_nelp.edxapp_wrapper.branding import get_visible_courses
+from eox_nelp.edxapp_wrapper.modulestore import modulestore
+from eox_nelp.edxapp_wrapper.site_configuration import configuration_helpers
+from eox_nelp.edxapp_wrapper.student import CourseAccessRole, CourseEnrollment
+from eox_nelp.stats.decorators import cache_method
+
+
+@cache_method
+def get_cached_courses(tenant):  # pylint: disable=unused-argument
+    """
+    Returns the visible courses. This method is just a wrapper of get_visible_courses that
+    allows to cache the result.
+
+    Args:
+        tenant<str>: String tenant identifier(site.domain)
+
+    Return:
+        list[<CourseOverview>]: List of CourseOverview records.
+    """
+    return get_visible_courses()
+
+
+@cache_method
+def get_course_metrics(course_key):
+    """
+    This allows to get the course stats metrics based on the course key.
+
+    Args:
+        course_key<opaque-key>: Course identifier.
+
+    Return:
+        <Dictionary>: Contains the course's metrics.
+    """
+    stats_settings = getattr(settings, "STATS_SETTINGS", {})
+    course = modulestore().get_course(course_key)
+    chapters = course.get_children()
+
+    sequentials = []
+
+    for chapter in chapters:
+        sequentials += chapter.get_children()
+
+    verticals = []
+
+    for sequential in sequentials:
+        verticals += sequential.get_children()
+
+    allowed_block_types = stats_settings.get("API_XBLOCK_TYPES", [])
+    components = {key: 0 for key in allowed_block_types}
+
+    for vertical in verticals:
+        for component in vertical.children:
+            if component.block_type in allowed_block_types:
+                components[component.block_type] = components.get(component.block_type, 0) + 1
+
+    instructors = CourseAccessRole.objects.filter(course_id=course_key).values('user').distinct().count()
+    learners = CourseEnrollment.objects.filter(
+        course=course_key,
+        user__is_staff=False,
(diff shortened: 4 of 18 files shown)


eox_nelp/stats/metrics.py
```diff
@@ -0,0 +1,136 @@
...
+
+    for chapter in chapters:
+        sequentials += chapter.get_children()
+
+    verticals = []
+
+    for sequential in sequentials:
+        verticals += sequential.get_children()
+
+    allowed_block_types = stats_settings.get("ALLOWED_VERTICAL_BLOCK_TYPES", [])
```
johanv26, line 59: Here I think is not vertical. Only block_types or xblocks, or component but I prefer only `xblock_types`. 

To analize a tenant I have the problem here that you must control the tenant xblocks installed, ok this is could be a general setting for sagar but this set that we have to managed(know) the installed xblocks. Something that is interesting, but what happen if a new release of openedx come with a new xblock. We have to add this one and add it to the setting?
Or this forces to query one time mongo and get all the xblock types and configure correctly the setting.🤔

eox_nelp/stats/metrics.py
```diff
@@ -0,0 +1,136 @@
...
+    components = {key: 0 for key in allowed_block_types}
```
johanv26, line 60: To analize a tenant I have the problem here that you must control the tenant xblocks installed, ok this is could be a general setting for sagar but this set that we have to managed(know) the installed xblocks. Something that is interesting, but what happen if a new release of openedx come with a new xblock. We have to add this one and add it to the setting?
Or this forces to query one time mongo and get all the xblock types and configure correctly the setting `ALLOWED_VERTICAL_BLOCK_TYPES`.🤔

eox_nelp/stats/decorators.py
```diff
@@ -0,0 +1,37 @@
...
+    Cache the function result to improve the response time.
+
+    Args:
+        func<function>: Target function to be cached.
+
+    Return:
+        <funtion>: Wrapper function.
+    """
+    def wrapper(*args, **kwargs):
+        key = f"{func.__name__}.{'-'.join(map(str, args))}.STATS_CACHE_KEY"
```
johanv26, line 21: I think that is better to relate more to the tenant the cache key, it could be other stats. `STATS_CACHE_KEY`  to `TENANT_STATS_CACHE_KEY`

johanv26:
I have a question related to the xblock allowed.
Mainly the endpoint is returning a good summary of course metrics. But my doubt is that this endpoint would give info about all the tenant. So return the section, subsection, units, and component(xblock_types)  but the count for all the courses of the tenant.
 This endpoint has it in a certain way but do I have to count or iterate the courses list?

eox_nelp/stats/decorators.py
andrey-canon: my question would be what are you talking about tenants here ? I mean we are in a multitenant context, however here we are talking about  a single method which functionality is just to cache the inner function result and that method shouldn't be aware of multi-tenant instances or something similar, therefore if we use   `TENANT_STATS_CACHE_KEY` that makes me think that the method has different caches from tenant  and that is false

andrey-canon:
> I have a question related to the xblock allowed. Mainly the endpoint is returning a good summary of course metrics. But my doubt is that this endpoint would give info about all the tenant. So return the section, subsection, units, and component(xblock_types) but the count for all the courses of the tenant. This endpoint has it in a certain way but do I have to count or iterate the courses list?

yes, you have to iterate

eox_nelp/stats/metrics.py
andrey-canon: you don't have to control any xblock at all, you have to control the API content that will be shown, so if you have a new xblock and you want to show that metric, just add it to the setting

eox_nelp/stats/api/v1/views.py
```diff
@@ -0,0 +1,68 @@
...
+    }
+    ```
+    """
+
+    def get(self, request):
+        """Return general tenant stats."""
+        tenant = request.site.domain
+
+        return Response({
+            "learners": metrics.get_learners_metric(tenant),
```
johanv26, line 65: Why if I dont have the videos and problems section if they should be by tenant.
image

Why for this keys video and problems I have to calculate them later.
https://github.com/nelc/frontend-essentials/pull/20/files#diff-683674ae0d4e56c5cc60a4c990b1dc7a78426e416462479e8bdb38f1fcff6909R105

eox_nelp/stats/api/v1/urls.py
```diff
@@ -0,0 +1,12 @@
...
+Course API URLs
+"""
+from django.urls import path
+
+from eox_nelp.stats.api.v1.views import GeneralTenantStatsView
+
+app_name = "eox_nelp"  # pylint: disable=invalid-name
+
+urlpatterns = [
+    path('tenant-stats/', GeneralTenantStatsView.as_view(), name="general-stats"),
```
johanv26, line 11: tenant

johanv26, line 11:
```suggestion
    path('tenant/', GeneralTenantStatsView.as_view(), name="general-stats"),
```

johanv26:
I was thinking about the approach of the endpoint and the stats API view behaviour. My opinion is that the section of courses has a  good logic but this would be a specific endpoint. For example, think in the course endpoints.
`/api/courses/v1/courses/`
`/eox-nelp/courses/v1/courses/`


They show the courses data for each tenant(tenant-aware). The endpoint is paginated and also you can access specific course id using `/api/courses/v1/courses/{course_id}`.

So keeping that in mind, I think this API would split in that way.
So the tenant route would give the stats summary (compact) of the tenant and the courses route would give all the  logic of
`get_course_metrics` https://github.com/eduNEXT/eox-nelp/pull/55/files#diff-43166e73e47980ce52493a5c95293b08351f1a03559ae6ad4dfcbbc8cfc54101R134

So the urls section of stats would be like
``` python
urlpatterns = [
    path('tenant/', GeneralTenantStatsView.as_view(), name="general-stats"),
    path('courses/', GeneralTenantStatsView.as_view(), name="courses-stats"),
    path('courses/{course_id}', GeneralTenantStatsView.as_view(), name="course-stats"),
]
```

andrey-canon:
> I was thinking about the approach of the endpoint and the stats API view behaviour. My opinion is that the section of courses has a good logic but this would be a specific endpoint. For example, think in the course endpoints. `/api/courses/v1/courses/` `/eox-nelp/courses/v1/courses/` 
> 
> They show the courses data for each tenant(tenant-aware). The endpoint is paginated and also you can access specific course id using `/api/courses/v1/courses/{course_id}`.
> 
> So keeping that in mind, I think this API would split in that way. So the tenant route would give the stats summary (compact) of the tenant and the courses route would give all the logic of `get_course_metrics` https://github.com/eduNEXT/eox-nelp/pull/55/files#diff-43166e73e47980ce52493a5c95293b08351f1a03559ae6ad4dfcbbc8cfc54101R134
> 
> So the urls section of stats would be like
> 
> ```python
> urlpatterns = [
>     path('tenant/', GeneralTenantStatsView.as_view(), name="general-stats"),
>     path('courses/', GeneralTenantStatsView.as_view(), name="courses-stats"),
>     path('courses/{course_id}', GeneralTenantStatsView.as_view(), name="course-stats"),
> ]
> ```

easy peasy

eox_nelp/stats/tests/api/v1/tests_views.py
```diff
@@ -0,0 +1,230 @@
...
+            "metrics": [
+                fake_metric, fake_metric, fake_metric, fake_metric
+            ],
+        }
+        url_endpoint = reverse("stats-api:v1:general-stats")
+
+        response = self.client.get(url_endpoint)
+
+        self.assertEqual(status.HTTP_200_OK, response.status_code)
+        self.assertEqual({"html": 20, "problem": 40, "video": 60}, response.data["components"])
```
johanv26, line 85: This could be something like
```suggestion
        self.assertEqual({"html": len(mock_metrics.get_courses_metrics())*fake_metric["html"], "problem": 40, "video": 60}, response.data["components"])
```
Also may be could be a dict comprehension
``` python
{key:value*len(mock_metrics.get_courses_metrics()) for (key, value) in fake_metric["components"].items()}
```

eox_nelp/stats/tests/api/v1/tests_views.py
```diff
@@ -0,0 +1,230 @@
...
+    @override_settings(MIDDLEWARE=["eox_tenant.middleware.CurrentSiteMiddleware"])
+    @patch("eox_nelp.stats.api.v1.views.metrics")
+    def test_get_list(self, mock_metrics):
+        """
+        Test a get request, this will verify the standard view behavior by checking the call of the metrics functions.
+
+        Expected behavior:
+            - Status code 200.
+            - the total_course key is present and has the expected value
+            - the lenght of metrics is the expected.
```
johanv26, line 123:
```suggestion
            - the length of metrics expected.
```

eox_nelp/stats/tests/api/v1/tests_views.py
```diff
@@ -0,0 +1,230 @@
...
+            "metrics": [
+                {}, {}, {}
+            ],
+        }
+        url_endpoint = reverse("stats-api:v1:courses-stats")
+
+        response = self.client.get(url_endpoint)
+
+        self.assertEqual(status.HTTP_200_OK, response.status_code)
+        self.assertEqual(3, response.data["total_courses"])
```
johanv26, line 136:
```suggestion
        self.assertEqual(mock_metrics.get_courses_metrics()["total_courses"], response.data["total_courses"])
```

eox_nelp/stats/tests/api/v1/tests_views.py
```diff
@@ -0,0 +1,230 @@
...
+        self.assertEqual(3, len(response.data["metrics"]))
```
johanv26, line 137:
```suggestion
        self.assertEqual(mock_metrics.get_courses_metrics()["metrics"], len(response.data["metrics"]))
```

eox_nelp/stats/tests/api/v1/tests_views.py
```diff
@@ -0,0 +1,230 @@
...
+            - get_cached_courses is called with the right parameter.
+            - filter is called with the right parameter.
+            - first is called once.
+            - get_course_metrics is not called.
+        """
+        course_id = "course-v1:potato+CS102+2023"
+        courses_mock = Mock()
+        courses_mock.filter.return_value.first.return_value = None
+        mock_metrics.get_cached_courses.return_value = courses_mock
+
```
johanv26, line 205: extra white space. `url_endpoint` should be one line above.

eox_nelp/stats/metrics.py
andrey-canon: I like that structure however if we just require the total, add a dictionary looks unnecessary and just add more complexity to the response, so do we need any extra information ?

eox_nelp/stats/metrics.py
```diff
@@ -0,0 +1,136 @@
...
+
+    Args:
+        tenant<str>: This argument is used as identifier by the cache_method decorator.
+
+    Return:
+        <int>: Total of instructors.
+    """
+    current_site_orgs = configuration_helpers.get_current_site_orgs()
+
+    return CourseAccessRole.objects.filter(org__in=current_site_orgs).values('user').distinct().count()
```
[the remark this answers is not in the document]
andrey-canon: yes. that must be counted

eox_nelp/stats/api/v1/views.py
```diff
@@ -0,0 +1,145 @@
...
+            "courses": courses.get("total_courses", 0),
+            "instructors": metrics.get_instructors_metric(tenant),
+            "components": components
+        })
+
+
+class GeneralCourseStatsView(APIView):
```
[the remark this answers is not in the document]
andrey-canon: ListView doesn't return the detail

eox_nelp/stats/tests/api/v1/tests_views.py
andrey-canon: indeed that should be `the length of metrics is as expected`

eox_nelp/stats/api/v1/views.py
```diff
@@ -0,0 +1,145 @@
...
+            "html": 133,
+            "openassessment": 0,
+            "problem": 49,
+            "video": 0
+        }
+    }
+    ```
+    """
+
+    def get(self, request):
```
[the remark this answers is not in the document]
andrey-canon: imo this doesn't deserve cache, the execution time is too low

eox_nelp/stats/tests/api/v1/tests_views.py
andrey-canon: when you executes a mock like `mock_metrics.get_courses_metrics()` you are increasing the number of calls  that affects some tests

eox_nelp/stats/metrics.py
johanv26: I don't like adding the dictionary but it reduces the good shape due to the  "uncertainty" of recovering the xblocks without knowing them. I mean this is an approach to avoid the setting of `ALLOWED_VERTICAL_BLOCK_TYPES`.

johanv26:
@andrey-canon are you going to use the json api renderer to nest the responses in the data field?
https://github.com/eduNEXT/eox-nelp/pull/55/commits/03f050f403f421d67d3349b8655a1ab5b44ec935#diff-8b60727e9925b9f343cdedd436aaad2bd26e9e7522fb94d6236960b2646d82d2R18

eox_nelp/stats/metrics.py
andrey-canon: That setting aims to keep the same kind components in the API response no matter the course content, so the API structure is constant in the same tenant, if I implement this I'm just bringing again the initial problem but in a nested way that is more complex. IMO the discussion shouldn't be how hide the inconsistent structure, if you don't like the setting purpose, tell me arguments of why we should implement an API that change its structure based on the course content

johanv26: I think that the reason is that is a stats API. Stats are based on the data of the app. In this case,  mongo. Mongo is a not relational database so you could set there whatever "kind" of documents. IMO it makes sense that the API should reflect the data stored in Mongo, and in this case, the API reflects the distinct cases of xblocks_types stored. Anyway, this discussion seems.
Course managers have often presented the question of which kind of xblocks we have. 
At the moment there are queries to reach it, but this seems like an opportunity to improve that.
This course's endpoint is not relevant to the tenant stats of the home page, so we could discuss this furthermore in the future.
https://3.basecamp.com/3966315/buckets/12732851/messages/4886008049

andrey-canon: In a good architecture the implementation shouldn't depend on the source of information, that could be mongo, mysql, postgres, .. etc, so the mongo argument doesn't fit here since this a different layer however you make me think that I should implement something stricter, probably in the future. Finally looks like you are thinking in something like GraphQL that is different to an API REST

eox_nelp/stats/tests/tests_metrics.py
```diff
@@ -0,0 +1,371 @@
...
+
+    def test_get_right_instructors_metric(self):
+        """Based on the initial conditions, this check that the course metrics has the expected instructors value.
+
+        Expected behavior:
+            - 'instructors' value is the expected
+        """
+        course = get_course_metrics(self.course_key)
+
+        self.assertEqual(5, course["instructors"])
```
johanv26, line 314: This one could be compared with this. https://github.com/eduNEXT/eox-nelp/pull/55/files#diff-ed3b504ce4a13370c20a7d619e584c45fee7c0d005ca4a6abc4974cb05ec1a04R262
but that var is used twice.
Like above  the rename of var could help.

eox_nelp/tests/utils.py
```diff
@@ -43,6 +43,8 @@ def set_key_values(element):
     for key, value in element.items():
         if isinstance(value, dict):
             setattr(model, key, set_key_values(value))
+        elif isinstance(value, list):
+            setattr(model, key, generate_list_mock_data(value))
```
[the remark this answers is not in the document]
johanv26: If you can add the list example in the docstring of the method xD.

andrey-canon:
@johanv26 did you finish your review ?

johanv26:
> @johanv26 did you finish your review ?

At this point, I have finished.

eox_nelp/stats/tests/tests_metrics.py
```diff
@@ -0,0 +1,371 @@
...
+        self.modulestore.return_value.get_course.return_value = course
+
+        # this block set the CourseEnrollment mock and its returned values.
+        filter_result = CourseEnrollment.objects.filter.return_value
+        values_result = filter_result.values.return_value
+        distinct_result = values_result.distinct.return_value
+        distinct_result.count.return_value = 5874
+
+        # this block set the CourseAccessRole mock and its returned values.
+        filter_result = CourseAccessRole.objects.filter.return_value
```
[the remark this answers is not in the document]
andrey-canon: That name doesn't matter

johanv26 (approved):
Squash and LGTM. The xblock discussion is for the future xD.

PR #57, opened 2023-06-29
feat: add tenant stats view

andrey-canon:
## Description

Add stats view issue https://edunext.atlassian.net/browse/FUTUREX-443
image


eox_nelp/urls.py
```diff
@@ -29,4 +29,5 @@
         include('eox_nelp.course_experience.frontend.urls', namespace='course-experience-frontend'),
     ),
     path('api/stats/', include('eox_nelp.stats.api.urls', namespace='stats-api')),
+    path('stats/', include('eox_nelp.stats.urls', namespace='stats')),
```
johanv26, line 32: Here we have to discuss which would be the default path. I like the prefix fronted cause as a user I relate that it has react or some js extra.
https://github.com/eduNEXT/eox-nelp/pull/57/files#diff-99d9cbe815961e1cc232cd5191394ac2e794e5ea823be05f5f9addf7331a39a7R28

eox_nelp/stats/urls.py
```diff
@@ -0,0 +1,10 @@
+"""frontend templates urls for course_experience"""
```
johanv26, line 1:
```suggestion
"""frontend templates urls for stats"""
```

eox_nelp/stats/views.py
```diff
@@ -0,0 +1,40 @@
...
+def get_tenant_stats(request):
+    """
+    Simple function based view that renders the StatsContainer essential component.
+    By default this show nothing since this requires the specific query para to show the content.
+
+    Examples:
+        by default the view render all componets: /eox-nelp/stats/tenant/
+
+        filter components: eg dont show videos and courses.
+        /eox-nelp/stats/tenant/?videos=False&courses=false
```
johanv26, line 21:
```suggestion
        /eox-nelp/stats/tenant/?videos=false&courses=false
```

johanv26 (approved):
I approve. I made a change but in another commit if necessary for future management. I gonna merge it.

johanv26:
My changes in summary have the frontend in the path and the positive logic.
image

eox_nelp/urls.py
andrey-canon: That's right the react components should be in the frontend and that's all, if I didn't include the view, inside the frontend folder, is  because a Django view is not a frontend element and that logic is not js or css and finally to add the path frontend in the url is redundant a view is an API or belongs to the presentation layer and you probably won't find  the path frontend in other views

PR #59, opened 2023-07-05
feat!: use explicit stats' query params

andrey-canon:
## Description

This implements explicit stats' query params since the current options are ambiguous  

## Testing instructions
1. check this branch
2.  test new queryparams, example `/eox-nelp/frontend/stats/tenant/?hideVideos=true&hideCourses&hideLearners=false`

### Result
image

### DIFF
--- a/eox_nelp/stats/templates/tenant_stats.html
+++ b/eox_nelp/stats/templates/tenant_stats.html
@@ -8,11 +8,11 @@
 <body>
     <div id="tenant-stats"></div>
     <script>
-      var showVideos = "true" === "${videos}";
-      var showCourses = "true" === "${courses}";
-      var showProblems = "true" === "${problems}";
-      var showLearners = "true" === "${learners}";
-      var showInstructors = "true" === "${instructors}";
+      var showVideos = "false" === "${hideVideos}";
+      var showCourses = "false" === "${hideCourses}";
+      var showProblems = "false" === "${hideProblems}";
+      var showLearners = "false" === "${hideLearners}";
+      var showInstructors = "false" === "${hideInstructors}";
     </script>
     <script src="${static.url('tenant_stats/js/tenant_stats.js')}"></script>
 </body>
--- a/eox_nelp/stats/tests/test_views.py
+++ b/eox_nelp/stats/tests/test_views.py
@@ -37,26 +37,26 @@ class GetTenantStatsTestCase(TestCase):
         self.assertEqual(self.template_name, response.templates[0].name)
         self.assertContains(response, '<div id="tenant-stats"></div')
 
-    @data("videos", "courses", "problems", "instructors", "learners")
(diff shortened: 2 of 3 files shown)


eox_nelp/stats/templates/tenant_stats.html
```diff
@@ -8,11 +8,11 @@
 <body>
     <div id="tenant-stats"></div>
     <script>
-      var showVideos = "true" === "${videos}";
-      var showCourses = "true" === "${courses}";
-      var showProblems = "true" === "${problems}";
-      var showLearners = "true" === "${learners}";
-      var showInstructors = "true" === "${instructors}";
+      var showVideos = "false" === "${hideVideos}";
```
johanv26, line 11: I like the refactoring of the query parameters in the theme. The only thing is the logic, so In control, we have to thing in hidden items. So I have to use a blacklist logic.
Mainly, I am looking at this view as the feedback API endpoint for example.  We have a boolean parameter `public`.
If we set it true it means public but if I set it false not public or "private, hide-public". The thing is that we want to manage this logic of blacklist? 
Also, I have a mismatch of the query param, this is a django view, query params in snake case, but ups the component use Camelcase.

eox_nelp/stats/templates/tenant_stats.html
```diff
@@ -8,11 +8,11 @@
...
+      var showCourses = "false" === "${hideCourses}";
+      var showProblems = "false" === "${hideProblems}";
+      var showLearners = "false" === "${hideLearners}";
```
johanv26, line 14: Here is that thing where we have to consider changing the names and matching connection with component, why if the component has ShowCourses we have to connect it with hideCourses? This could be confusing.
[😶‍🌫️](https://emojipedia.org/face-in-clouds/)

andrey-canon: this comment is weird since you removed the thing that you are proposing

johanv26: Actually, I changed it to true in my first approach. Then due to my confusion about snake_case and camel_case, I changed to videos, courses, problems.learners.. they all fit snake and camel case

eox_nelp/stats/templates/tenant_stats.html
andrey-canon: please stop using racist terms, but  anyway that is just a semantic thing if the label say hide or show the user should be smart enough to select the right option

eox_nelp/stats/templates/tenant_stats.html
andrey-canon: yes, I checked again your changes and you didn't  proposed any change in the way how the api is consumed, you only changed the default values to show everything  but using confusing names

andrey-canon:
closed in favor of https://github.com/eduNEXT/eox-nelp/pull/60

PR #61, opened 2023-07-05
feat: use query params in snake case

andrey-canon:
## Description

This changes the format of the query params from camel case to snake case

### DIFF
--- a/eox_nelp/stats/views.py
+++ b/eox_nelp/stats/views.py
@@ -15,25 +15,25 @@ def get_tenant_stats(request):
     By default this show nothing since this requires the specific query para to show the content.
 
     Examples:
-        renders just video card /eox-nelp/stats/tenant/?showVideos=true
+        renders just video card /eox-nelp/stats/tenant/?show_videos=true
 
         render multiple components
-        /eox-nelp/stats/tenant/?showVideos=true&showCourses=true&showInstructors=true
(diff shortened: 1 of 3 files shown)


eox_nelp/stats/views.py
```diff
@@ -15,25 +15,25 @@ def get_tenant_stats(request):
...
-        "showCourses": "false",
-        "showVideos": "false",
-        "showProblems": "false",
-        "showLearners": "false",
-        "showInstructors": "false",
+        "show_courses": "false",
+        "show_videos": "false",
+        "show_problems": "false",
+        "show_learners": "false",
+        "show_instructors": "false",
```
johanv26, line 36:
```suggestion
        "show_courses": "true",
        "show_videos": "true",
        "show_problems": "true",
        "show_learners": "true",
        "show_instructors": "true",
```
This would be true to render all if no query params are configured.

andrey-canon: please read the pr description

johanv26: So the true logic would be other PR?

andrey-canon: yes, a different pr in that way we can focus in small changes

johanv26 (approved):
I approve of the kind reminder to move to positive logic. To the API view, render all the components without query params.

PR #74, opened 2023-08-02
refactor: implementing abstract API class

andrey-canon:
## Description
This move the general logic from the FuturexAPIClient to a new Abstract class.

### DIFF
--- a/eox_nelp/api_clients/__init__.py
+++ b/eox_nelp/api_clients/__init__.py
@@ -1,0 +1,129 @@
+"""This file contains the common functions and classes for the api_clients module.
+
+Classes:
+    AbstractApiClient: Base API class.
+"""
+import logging
+from abc import ABC, abstractmethod
+
+import requests
+from django.core.cache import cache
+from oauthlib.oauth2 import BackendApplicationClient
+from requests_oauthlib import OAuth2Session
+from rest_framework import status
+
+LOGGER = logging.getLogger(__name__)
+
+
+class AbstractApiClient(ABC):
+    """Abstract api client class, this implement a basic authentication method and defines methods POST and GET"""
+
+    @property
+    @abstractmethod
+    def base_url(self):
+        """Abstract base_url property method."""
+        raise NotImplementedError
+
+    def __init__(self, client_id, client_secret):
+        """
+        Abstract ApiClient creator, this will set the session based on the authenticate result.
+
+        Args:
+            client_id<str>: Public application identifier.
+            client_secret<str>: Confidential identifier.
+        """
+        key = f"{client_id}-{client_secret}"
+        headers = cache.get(key)
+
+        if not headers:
+            response = self._authenticate(client_id, client_secret)
+            headers = {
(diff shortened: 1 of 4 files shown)


eox_nelp/api_clients/tests/tests_futurex.py
```diff
@@ -1,182 +1,22 @@
 """This file contains all the test for futurex api client file.
+
 Classes:
-    LikeDislikeUnitExperienceTestCase: Test LikeDislikeUnitExperienceView.
```
johanv26, line 3: Nice catch

eox_nelp/api_clients/tests/tests_futurex.py
```diff
@@ -1,182 +1,22 @@
...
-        requests_mock.Session.return_value.get.assert_called_with(
-            url=f"{api_client.base_url}/fake/path",
-            params=params,
-        )
-        self.assertEqual(logs.output, [
-            f"ERROR:{futurex.__name__}:{log_error}"
-        ])
+    def setUp(self):
+        """Setup common conditions for every test case"""
+        self.api_class = FuturexApiClient
```
johanv26, line 19: These lines could be?
https://github.com/eduNEXT/eox-nelp/blob/1e501f2b3b74c966c766675cc5d292aeebfbcbd4/eox_nelp/api_clients/tests/tests_futurex.py#L44
https://github.com/eduNEXT/eox-nelp/blob/1e501f2b3b74c966c766675cc5d292aeebfbcbd4/eox_nelp/api_clients/tests/tests_futurex.py#L64
``` python
api_client = self.api_client
```

eox_nelp/api_clients/futurex.py
```diff
@@ -25,100 +18,14 @@ class FuturexApiClient:
...
-            "An error has occurred trying to make a get request to %s with status code %s",
-            url,
-            response.status_code,
-        )
+        super().__init__(client_id, client_secret)
 
-        return {}
+    @property
+    def base_url(self):
+        return getattr(settings, "FUTUREX_API_URL")
```
johanv26, line 28: This needs default value?

eox_nelp/api_clients/__init__.py
```diff
@@ -0,0 +1,123 @@
...
+        if response.status_code == status.HTTP_200_OK:
+            return response.json()
+
+        LOGGER.error(
+            "An error has occurred trying to make a get request to %s with status code %s",
+            url,
+            response.status_code,
+        )
+
+        return {}
```
johanv26, line 123: I know this is a refactor, but here you could a message of error or something related to the response...
How can I know if it is a 200  empty or an error? By the way, is the opportunity to handle it or maybe in the future because is working with that way now.

eox_nelp/api_clients/futurex.py
andrey-canon: No, since in the past the decision was if someone wants to use it they must set up everything otherwise this would raise an exception

johanv26 (approved):

PR #75, opened 2023-08-02
feat: add certificates api client

andrey-canon:
## Description
This is a first version if the certificates APIclient ,however this is just an initial approach since that must be improved when we have more details about the API.
[issue](https://edunext.atlassian.net/browse/FUTUREX-478)

### DIFF
--- a/eox_nelp/api_clients/certificates.py
+++ b/eox_nelp/api_clients/certificates.py
@@ -1,0 +1,68 @@
+"""Client module for external certificate's API integration.
+
+Classes:
+    ExternalCertificatesApiClient: Class to interact with NELP external certificates service.
+"""
+from django.conf import settings
+
+from eox_nelp.api_clients import AbstractApiClient
+
+
+class ExternalCertificatesApiClient(AbstractApiClient):
+    """Allow to perform multiple external certificates operations."""
+
+    def __init__(self):
+        client_id = getattr(settings, "EXTERNAL_CERTIFICATES_API_CLIENT_ID")
+        client_secret = getattr(settings, "EXTERNAL_CERTIFICATES_API_CLIENT_SECRET")
+
+        super().__init__(client_id, client_secret)
+
+    @property
+    def base_url(self):
+        return getattr(settings, "EXTERNAL_CERTIFICATES_API_URL")
+
+    def create_external_certificate(self, certificate_data):
+        """This will create an external certificate based on the input data, this data should have the
+        following keys:
+
+            id <mandatory>: edx-platform certificate identifier.
+            created_at <mandatory>: when the certificate was created.
+            expiration_date <mandatory>: when the certificate expires.
+            grade <mandatory>: The associated grade with the certificate.
+            is_passing <mandatory>: Boolean value that represent if the user has passed the course.
+            user <mandatory>: Dictionary with the following data:
+                national_id: User National identifier.
+                englishs_name <optional>: User name in English.
+                arabic_name <optional>: User name in Arabic.
+        Args:
+            certificate_data<Dict>: Information about a the certificate.
+
+        Returns:
+            response<Dict>: requests response as dictionary.
+
+        Raise:
+            KeyError: This will be raised when the mandatory are excluded in the certificate data.
+        """
+        path = "certificates"  # This is not clear at all
+        user = certificate_data["user"]
+        payload = {
+            "reference_id": certificate_data["id"],
+            "date": {
+                "issuance": str(certificate_data["created_at"]),
+                "expiration": str(certificate_data["expiration_date"]),
+            },
+            "individual": {
+                "name_en": user.get("english_name", ""),
+                "name_ar": user.get("arabic_name", ""),
+                "id": user["national_id"],
+                "id_type": "saudi",
+            },
+            "group_code": "fail",  # This is not clear
+            "certificate_type": "completion",  # What types do we have ?
+            "metadata": {
+                "degree": certificate_data["grade"],
+                "FAIL": certificate_data["is_passing"],
+            }
+        }
+
+        return self.make_post(path, payload)
--- a/eox_nelp/api_clients/tests/tests_certificates.py
+++ b/eox_nelp/api_clients/tests/tests_certificates.py
@@ -1,0 +1,65 @@
+"""This file contains all the test for certificates api client file.
+
+Classes:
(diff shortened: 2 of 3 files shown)


eox_nelp/api_clients/certificates.py
```diff
@@ -0,0 +1,68 @@
...
+        path = "certificates"  # This is not clear at all
+        user = certificate_data["user"]
+        payload = {
+            "reference_id": certificate_data["id"],
+            "date": {
+                "issuance": str(certificate_data["created_at"]),
+                "expiration": str(certificate_data["expiration_date"]),
+            },
+            "individual": {
+                "name_en": user.get("english_name", ""),
```
johanv26, line 55: I have a question here.
In this section user is a dict and not a User object?

eox_nelp/api_clients/certificates.py
```diff
@@ -0,0 +1,68 @@
...
+from django.conf import settings
+
+from eox_nelp.api_clients import AbstractApiClient
+
+
+class ExternalCertificatesApiClient(AbstractApiClient):
+    """Allow to perform multiple external certificates operations."""
+
+    def __init__(self):
+        client_id = getattr(settings, "EXTERNAL_CERTIFICATES_API_CLIENT_ID")
```
johanv26, line 15: This work, if the auth, is made like Django client credentials body way.
If this change, we have to adapt our AbstractApiClient to work like that or add another method to authenticate....
image
I say this because at this moment we don't know what is the way of auth.
It could be different ways like
image

eox_nelp/api_clients/tests/tests_certificates.py
```diff
@@ -0,0 +1,65 @@
...
+        """Test when the mandatory fields has not been sent.
+
+        Expected behavior:
+            - Raise KeyError exception.
+        """
+        auth_mock.return_value = {}
+        data = {}
+        api_client = ExternalCertificatesApiClient()
+
+        self.assertRaises(KeyError, api_client.create_external_certificate, data)
```
johanv26, line 65: If the auth is a mock, the API client first raised is a key error of dict?

eox_nelp/api_clients/tests/tests_certificates.py
```diff
@@ -0,0 +1,65 @@
...
+    @patch.object(ExternalCertificatesApiClient, "_authenticate")
+    def test_create_certificate(self, auth_mock, post_mock):
+        """Test successful post request.
+
+        Expected behavior:
+            - Response is the expected value
+        """
+        auth_mock.return_value = {}
+        expected_value = {
+            "status": {"success": True, "message": "successful", "code": 1}
```
johanv26, line 32: I assume this is an assumption of the response of auth endpoint? xD heheeh

eox_nelp/api_clients/certificates.py
andrey-canon: https://github.com/eduNEXT/eox-nelp/blob/d12faa2b85365693fe1933d0830b090cca868dd5/eox_nelp/api_clients/certificates.py#L33

eox_nelp/api_clients/certificates.py
```diff
@@ -0,0 +1,68 @@
...
+
+        Returns:
+            response<Dict>: requests response as dictionary.
+
+        Raise:
+            KeyError: This will be raised when the mandatory are excluded in the certificate data.
+        """
+        path = "certificates"  # This is not clear at all
+        user = certificate_data["user"]
+        payload = {
```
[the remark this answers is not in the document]
andrey-canon: why body and not payload ?  why payload and not data ?
https://requests.readthedocs.io/en/latest/user/quickstart/#more-complicated-post-requests

eox_nelp/api_clients/certificates.py
andrey-canon: yes, since we don't know we should implement the standard way

eox_nelp/api_clients/tests/tests_certificates.py
andrey-canon: This really doesn't matter, this is the output of a mock  therefore that could be whatever you want, the important test here is that the output of `make_post` will be the output of the `create_external_certificate`

eox_nelp/api_clients/tests/tests_certificates.py
andrey-canon: probably I dont understand your question the auth_mock helps to skip errors when the client is initialized  line 63 and this test is for the execution of `create_external_certificate`

eox_nelp/api_clients/certificates.py
johanv26: Is strange, mainly time we manage we use the User Object,

eox_nelp/api_clients/tests/tests_certificates.py
johanv26: Sure, is strange that the auth is also mocked. Then I viewed the successful test and the auth is mocked too.

johanv26 (approved):

eox_nelp/api_clients/certificates.py
andrey-canon: this is a new function therefore we have to define the signature, in this case there is only one input data and that is a dictionary, so imo to include models inside the dictionary will be strange since it is not an standard  python type  

Indeed here  my concern  was where should I include the transformation data, since the original data is a  [CertificateData](https://github.com/eduNEXT/openedx-events/blob/main/openedx_events/learning/data.py#L100) and we must send a different format(API format), please check this pr https://github.com/eduNEXT/eox-nelp/pull/76, and let me know your thoughts  about the general implementation

eox_nelp/api_clients/tests/tests_certificates.py
andrey-canon: The authentication happens when the class is initialized  if the initialization fails the class instance doesn't exists therefore the method can not be called  and these tests can not be performed, and keep in mind these tests are not related with the authentication

eox_nelp/api_clients/certificates.py
johanv26: That has total logic due to delegating async tasks we cant send objects in args, but we can send dicts.
So is like serializing the user object into dict.
https://github.com/eduNEXT/eox-nelp/pull/76/files#diff-902ce53d9739afb0b3d31006e61041cc409b76c1a250609145a838a8d070bf71R192

PR #77, opened 2023-08-04
feat: add api client based on basic authentication.

andrey-canon:
## Description

This is part of https://edunext.atlassian.net/browse/FUTUREX-478, basically this includes a new feature that is to use an API with basic authentication and makes a small refactor in order to have classes with different kind of authentication

### DIFF
--- a/eox_nelp/api_clients/__init__.py
+++ b/eox_nelp/api_clients/__init__.py
@@ -7,8 +7,10 @@ import logging
 from abc import ABC, abstractmethod
 
 import requests
+from django.conf import settings
 from django.core.cache import cache
 from oauthlib.oauth2 import BackendApplicationClient
+from requests.auth import HTTPBasicAuth
 from requests_oauthlib import OAuth2Session
 from rest_framework import status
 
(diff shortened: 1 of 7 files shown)


eox_nelp/api_clients/tests/__init__.py
```diff
@@ -177,3 +141,80 @@ def test_failed_get(self, auth_mock, requests_mock):
...
+            - fetch_token was called with the right values.
+        """
+        fetch_token_mock = Mock()
+        fetch_token_mock.return_value = {
+            "token_type": "Bearer",
+            "access_token": "12345678abc",
+            "expires_in": 200,
+        }
+        oauth2_session_mock.return_value.fetch_token = fetch_token_mock
+        authentication_url = f"{settings.FUTUREX_API_URL}/oauth/token"
```
johanv26, line 176:
```suggestion
        authentication_url = f"{settings.FUTUREX_API_URL}/oauth2/token"
```
?

eox_nelp/api_clients/tests/__init__.py
```diff
@@ -177,3 +141,80 @@ def test_failed_get(self, auth_mock, requests_mock):
...
+        """Test case when the authentication response is valid.
+
+        Expected behavior:
+            - Session is set
+            - Session headers contains Authorization key.
+            - fetch_token was called with the right values.
+        """
+        fetch_token_mock = Mock()
+        fetch_token_mock.return_value = {
+            "token_type": "Bearer",
```
johanv26, line 171: I was checking and this auth only work with bearer. In the future could be considered the jwt way or possibility.

eox_nelp/api_clients/__init__.py
```diff
@@ -127,3 +96,98 @@ def make_get(self, path, payload):
...
+            response = oauth.fetch_token(
+                token_url=authenticate_url,
+                client_secret=self.client_secret,
+                include_client_id=True,
+            )
+            headers = {
+                "Authorization": f"{response.get('token_type')} {response.get('access_token')}"
+            }
+
+            cache.set(key, headers, response.get("expires_in", 300))
```
johanv26, line 151: Cache would only work for oauth2 and not basic auth?

eox_nelp/api_clients/tests/__init__.py
andrey-canon: I found [here](https://github.com/eduNEXT/eox-nelp/blob/3efc7316a8a0b66a08ec5b73264c0d0ba0070401/eox_nelp/api_clients/__init__.py#L148), that the kind of token depends on the response

eox_nelp/api_clients/tests/__init__.py
andrey-canon: from the doc

image

https://drive.google.com/file/d/16PXRFCFZuRQMjleQqkm_qSKSfddpwT8g/view

eox_nelp/api_clients/__init__.py
andrey-canon: yes, the cache is just for oauth2  and the main reason is that both systems work different, both systems  send the authentication data in the `Authorization` header however the authentication data is generated in a different way, oauth2 uses a client_id and a client_secret to make a request to generate a token, so the backend is on charge of the generation, on the other hand, basic  authorization doesn't need to make a request since the authorization data  is just the `username:password`  in base64

eox_nelp/api_clients/tests/__init__.py
johanv26: In edx-platform you can ask the type. By the way, is ok.
image

johanv26 (approved):

eox_nelp/api_clients/tests/__init__.py
andrey-canon: keep in mind that this was initially designed  to consume the futurex service and probably won't fit in different different situations

PR #83, opened 2023-08-10
feat: add prefix to reference_id parameter in certificate api client

andrey-canon:
## Description

This adds a prefix to the payload reference id parameter, I didn't include the host since I consider that information is not relevant because the certificate is not associated with a tenant, that is associated with a course and a course could be associated with multiple tenants

### DIFF
--- a/eox_nelp/api_clients/certificates.py
+++ b/eox_nelp/api_clients/certificates.py
@@ -49,7 +49,7 @@ class ExternalCertificatesApiClient(AbstractBasicAuthApiClient):
         path = "Certificates"
         user = certificate_data["user"]
         payload = {
-            "reference_id": certificate_data["id"],
+            "reference_id": f"nelc-openedx-lms-cert-{certificate_data['id']}",
             "date": {
                 "issuance": certificate_data["created_at"],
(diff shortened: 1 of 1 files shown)


eox_nelp/api_clients/certificates.py
```diff
@@ -49,7 +49,7 @@ def create_external_certificate(self, certificate_data):
         path = "Certificates"
         user = certificate_data["user"]
         payload = {
-            "reference_id": certificate_data["id"],
+            "reference_id": f"openedx-lms-cert-{certificate_data['id']}",
```
johanv26, line 52: I think that they could have an edx integration. Is something related futurex or nelp. Or just with the open is different than with edx. Also to remind the possibility of the existence of other openedx integration...?

johanv26 (approved):

eox_nelp/api_clients/certificates.py
andrey-canon: IMO the probability of that is almost null and probably when they have another integration you will have another job, but anyway I have already applied that

johanv26 (approved):

PR #96, opened 2023-09-12
Jlc/certificates metrics

johanv26:
## Description

Add certificates key for the metric backend.
This add a certificates key for each course metric with this shape for course endpoint
``` json
{
            "certificates": {
                "verified": {"downloadable": 0, "notpassing": 0},
                "honor": {"downloadable": 0, "notpassing": 0},
                "audit": {"downloadable": 0, "notpassing": 0},
                "professional": {"downloadable": 0, "notpassing": 0},
                "no-id-professional": {"downloadable": 5, "notpassing": 4},
                "masters": {"downloadable": 0, "notpassing": 0},
                "executive-education": {"downloadable": 0, "notpassing": 0},
                "paid-executive-education": {"downloadable": 0, "notpassing": 0},
                "paid-bootcamp": {"downloadable": 0, "notpassing": 0},
                "total": 9,
            }
}
```
For the tenant endpoint only the total certificates of the tenant is shown:
``` json
{
    "learners": 1,
    "courses": 3,
    "instructors": 2,
    "components": {
        "discussion": 0,
        "drag-and-drop-v2": 0,
        "html": 133,
        "openassessment": 0,
        "problem": 49,
        "video": 0
    },
    "certiticates": 12
}
```
## Testing instructions

Clone this branch and check the behavior of the stats endpoints.
```
/eox-nelp/api/stats/v1/tenant/
/eox-nelp/api/stats/v1/courses/
/eox-nelp/api/stats/v1/courses/course-v1:edX+cd202+2023
```
### After
Peek 2023-09-12 14-18

## Additional information

Jira story:
https://edunext.atlassian.net/browse/FUTUREX-489
## Checklist for Merge

- [ ] Updated documentation
- [ ] Rebased master/main
- [ ] Squashed commits

### DIFF
--- a/eox_nelp/migrations/0001_initial.py
+++ b/eox_nelp/migrations/0001_initial.py
@@ -95,6 +95,17 @@ class Migration(migrations.Migration):
                 ('org', models.CharField(blank=True, max_length=500, null=True))
             ],
         ),
+        migrations.CreateModel(
+            name='GeneratedCertificate',
+            fields=[
+                ('id', models.AutoField(verbose_name='ID', serialize=False, auto_created=True, primary_key=True)),
+                ('user', models.ForeignKey(to=settings.AUTH_USER_MODEL, on_delete=models.CASCADE)),
+                ('course_id', opaque_keys.edx.django.models.CourseKeyField(default=None, max_length=255, blank=True)),
+                ('grade', models.CharField(default='', max_length=5, blank=True)),
+                ('status', models.CharField(default='unavailable', max_length=32)),
+                ('mode', models.CharField(default='honor', max_length=32, choices=[('verified', 'verified'), ('honor', 'honor'), ('audit', 'audit'), ('professional', 'professional'), ('no-id-professional', 'no-id-professional')])),
+            ],
+        ),
     ]
 
     if getattr(settings, 'TESTING_MIGRATIONS', False):
--- a/eox_nelp/stats/metrics.py
+++ b/eox_nelp/stats/metrics.py
@@ -6,14 +6,18 @@ functions:
     get_learners_metric: Return number of learners, for the visible courses.
     get_instructors_metric: Return number of instructors, for the visible courses.
     get_courses_metrics: Return metrics for the visible courses.
+    get_course_certificates_metric: Return a dict metric representin the certificates of a course.
 """
 from django.conf import settings
+from eox_core.edxapp_wrapper.certificates import get_generated_certificate
 
 from eox_nelp.edxapp_wrapper.branding import get_visible_courses
 from eox_nelp.edxapp_wrapper.modulestore import modulestore
 from eox_nelp.edxapp_wrapper.site_configuration import configuration_helpers
 from eox_nelp.edxapp_wrapper.student import CourseAccessRole, CourseEnrollment
 from eox_nelp.stats.decorators import cache_method
+
+GeneratedCertificate = get_generated_certificate()
 
 
 @cache_method
@@ -70,6 +74,7 @@ def get_course_metrics(course_key):
         user__is_staff=False,
         user__is_superuser=False
     ).values('user').distinct().count()
+    certificates = get_course_certificates_metric(course_key)
 
     return {
         "id": str(course_key),
@@ -79,7 +84,8 @@ def get_course_metrics(course_key):
         "sections": len(chapters),
         "sub_sections": len(sequentials),
         "units": len(verticals),
-        "components": components
+        "components": components,
+        "certificates": certificates,
     }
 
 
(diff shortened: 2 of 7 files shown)


eox_nelp/stats/metrics.py
```diff
@@ -134,3 +140,49 @@ def get_courses_metrics(tenant):
...
+                "professional": {...},
+                "no-id-professional": {
+                    "downloadable": 5,
+                    "notpassing": 4,
+                },
+                "masters": {...},
+                "executive-education": {...}
+                "paid-executive-education":{...},
+                "paid-bootcamp": {...},
+                "total": 0
```
andrey-canon, line 167: since you are splitting the certificates based on their status the total metric should reflect also that division

eox_nelp/migrations/0001_initial.py
```diff
@@ -95,6 +95,26 @@ class Migration(migrations.Migration):
                 ('org', models.CharField(blank=True, max_length=500, null=True))
             ],
         ),
+        migrations.CreateModel(
```
andrey-canon, line 98: what you are creating a a backend if you are creating the model ?

johanv26: Because the backend allows me to load the class that passes the data to the test db.
https://github.com/eduNEXT/eox-nelp/blob/f398520f00a55ef12283d7563ee2624b49dbf2b6/eox_nelp/stats/tests/tests_metrics.py#L273
https://github.com/eduNEXT/eox-nelp/blob/f398520f00a55ef12283d7563ee2624b49dbf2b6/eox_nelp/edxapp_wrapper/test_backends/certificates_m_v1.py#L32-L58

eox_nelp/stats/metrics.py
johanv26: Could you explain this one a little more, I mean I am setting the total without the division. 
So the division would reflect something like the total: 
``` python 
"total" : {
"no-id-professional": 5,
"audit": 5
}
```
Is that what you say?
Because if is that I prefer to set the total of each one inside each key?? But I didn't find it necessary due you only have to sum the field of each one

eox_nelp/migrations/0001_initial.py
andrey-canon: generally we don't duplicate models from edx-platform, why this case is so especial ?

eox_nelp/stats/metrics.py
andrey-canon: I didn't say mode 
```
                "total": {
                    "downloadable": 5,
                    "notpassing": 4,
                },
```

eox_nelp/migrations/0001_initial.py
johanv26: Actually is because in the test I use the model to generate the certificates in `db`. So the mainly reason is that, as you can see there course_overview also was recreated, but with less fields
https://github.com/eduNEXT/eox-nelp/blob/f398520f00a55ef12283d7563ee2624b49dbf2b6/eox_nelp/migrations/0001_initial.py#L83-L97
The reasons is that I want to test against like "`real certs data"`...

johanv26: So in the test, as you can see I don't mock any function of the certificates because the django test db has the model and data. So functions like GeneratedCertficates.objects... work and return data from test db.https://github.com/eduNEXT/eox-nelp/blob/f398520f00a55ef12283d7563ee2624b49dbf2b6/eox_nelp/stats/tests/tests_metrics.py#L273-L278

andrey-canon: Course overview was created because another model use that as foreign key, there was no another option

IMO there are another alternatives like Mock that could fit here without adding fields that you even don't use

johanv26: Actually, I started trying to use the mock approach, therefore as I used different Django queryset attributes and mixed them, I found that I have to do a lot of different mocks approaches and also with sense. So I stated to be a little mixed, so finally, I preferred to load the model and get data as the django queryset would work.
https://github.com/eduNEXT/eox-nelp/blob/f398520f00a55ef12283d7563ee2624b49dbf2b6/eox_nelp/stats/tests/tests_metrics.py#L257-L269

Also I set another fields, maybe extra but that could be useful if others created a certificate in the django db test model.

Actually all was loaded from here. Practically   I have the same model that the openedx
https://github.com/openedx/edx-platform/blob/master/lms/djangoapps/certificates/models.py#L224-L240

andrey-canon: That's the point we can not copy all the models that openedx use

eox_nelp/stats/metrics.py
johanv26: > I didn't say mode
> 
> ```
>                 "total": {
>                     "downloadable": 5,
>                     "notpassing": 4,
>                 },
> ```

@andrey-canon  what about this approach ?
Peek 2023-09-12 17-55

andrey-canon: did you remove the modes ?

andrey-canon (approved):

PR #100, opened 2023-10-09
feat: add custom social_details pipeline

andrey-canon:
## Description

This is an extension of `social_core.pipeline.social_auth.social_details` that allows to map response fields in standard user fields.

Part of https://edunext.atlassian.net/browse/FUTUREX-515

## Testing instructions

I tested this by adding the file to the debug server on production 
This change was required 
`AUTHENTICATION_BACKENDS.insert(0, "eox_core.social_tpa_backends.ConfigurableOpenIdConnectAuth")`

## Result 
sso-mt

### DIFF
--- a/eox_nelp/auth_pipeline.py
+++ b/eox_nelp/auth_pipeline.py
@@ -1,0 +1,30 @@
+"""Custom NELP  authentication pipelines
+
+functions:
+    social_details: Allows to map response fields to user standar fields.
+"""
+from django.conf import settings
+from social_core.pipeline.social_auth import social_details as social_core_details
+
+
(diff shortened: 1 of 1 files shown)


OmarIthawi (approved):
Looks great and generic!! Thanks!!

johanv26 (approved):
I like it. I think this could be useful to eox-core and tis 3partyauth backends, configurations...

OmarIthawi:
> I like it. I think this could be useful to eox-core and tis 3partyauth backends, configurations...

I agree. It seems like a great addition to the eox-core OpenID Connect backend.

PR #104, opened 2023-10-27
feat: add invalidate current user pipeline

andrey-canon:
## Description

This pipeline sets to None the current user in order to avoid invalid associations.
    This was implemented due to an unexpected behavior when a user is logged and a different
    user tries to authenticate by using a SAML IDP in the same browser, the result is that
    instead of having two different accounts the second user is associated to the first user's
    account, this behavior is the result of SESSION_COOKIE_SAMESITE = "None" however the
    possible fix, SESSION_COOKIE_SAMESITE = "Lax", has not been tested in the whole platform
    therefore that can not be implemented yet.


eox_nelp/third_party_auth/pipeline.py
```diff
@@ -28,3 +30,28 @@ def social_details(backend, details, response, *args, **kwargs):
         details["details"][key] = response.get(value)
 
     return details
+
+
+def invalidate_current_user(request, *args, user=None, **kwargs):  # pylint: disable=unused-argument
+    """This pipeline sets to None the current user in order to avoid invalid associations.
```
johanv26, line 36:
```suggestion
    """This pipeline sets to None the current user and logout request in order to avoid invalid associations.
```

eox_nelp/third_party_auth/pipeline.py
```diff
@@ -28,3 +30,28 @@ def social_details(backend, details, response, *args, **kwargs):
         details["details"][key] = response.get(value)
 
     return details
+
+
+def invalidate_current_user(request, *args, user=None, **kwargs):  # pylint: disable=unused-argument
```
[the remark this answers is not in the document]
andrey-canon: > What do u think of

logout functionality is not the main purpose of this, that is an extra functionality  of the pipeline,  the main purpose is to set the user to None, however that made me think that I should split this method  in order to keep just one responsibility per method

andrey-canon:
@johanv26 

https://github.com/eduNEXT/eox-nelp/assets/36200299/f67b1c8c-67fe-405b-8ac9-ec7957df277a

eox_nelp/third_party_auth/pipeline.py
```diff
@@ -0,0 +1,67 @@
...
+    """
+    if user:
+        return {
+            "user": None
+        }
+
+    return {}
+
+
+def close_active_session(request, *args, user=None, **kwargs):  # pylint: disable=unused-argument
```
johanv26, line 58: I have some thoughts about the name because `close_active_session` is a name that if I put it in the pipeline I ensure that this would close the actual session or do the logout even if the dict has user or not.

andrey-canon: sorry I didn't understand your concern, could you please give more details

johanv26: I mean that the name of the function `close_active_session` cover a lot. So If I added in the pipeline, I ensured the logout would happen without if statement `request.user != user`
Maybe `close_previous_session_mismatch_user` could be another name. Is only an opinion.

andrey-canon: if I understood since the name is `close_active_session` that should close all the session not matter the user, in this case the pipeline only closes the session if the current user is different from the session user, so what do you think of  `close_mismatch_session`  or `close_invalid_session` ??

johanv26: I like close_mismatch_session

johanv26 (approved):

PR #105, opened 2023-10-30
Jlc/course run course creator permissions

johanv26:
## Description

Modification for the  `{studio_url}/eox-nelp/api/v1/course_runs/ `access options.
Added 2 different permission to the view: (NelpCourseRunPermission, IsAuthenticated)
The  custom nelp permision for this view is related to needs.
This permission is based on the premise that the get, list, and update [forms ](https://github.com/eduNEXT/edunext-platform/blob/ednx-release/mango.master.nelp/cms/djangoapps/api/v1/views/course_runs.py#L29) are already verified by the following methods:

- https://github.com/eduNEXT/edunext-platform/blob/ednx-release/mango.master.nelp/common/djangoapps/student/auth.py#L128
-  https://github.com/eduNEXT/edunext-platform/blob/ednx-release/mango.master.nelp/common/djangoapps/student/auth.py#L75

So this permission is always True except for the create form.
For the POST method, there is some extra logic for the management.
 Only Global staff(admin_user) or OrgInstructorRole for the suggested org of the data could be created.
This changes the StafforgRole premise because staff would not have permission to edit according to this [line:](https://github.com/eduNEXT/edunext-platform/blob/ednx-release/mango.master.nelp/common/djangoapps/student/auth.py#L95
)

### Before
- The endpoint was accessible only with Django staff permission.
- Post requests only with Django staff permissions.

### After
- Endpoint accesible related to the user authenticated(jwt and session) and comparing the CourseAccessRole
- Post accesible for django staff and org course instructors.(create courses)
### Jira story
https://edunext.atlassian.net/browse/FUTUREX-616?atlOrigin=eyJpIjoiZGM0OTljNmYyOTMyNDNjMGIyNWU5ZTVhMmY1YzVjZTQiLCJwIjoiaiJ9


eox_nelp/cms/api/v1/permissions.py
```diff
@@ -0,0 +1,32 @@
...
+    """
+
+    def has_permission(self, request, view):
+        """
+        Define custom permission of Nelp case.
+        """
+        if request.method == "POST":
+            org = request.data.get("org", None)
+            return bool(
+                roles.GlobalStaff().has_user(request.user) or roles.OrgInstructorRole(org=org).has_user(request.user)
```
andrey-canon, line 29:
```suggestion
    def has_permission(self, request, view):
        """
        Define custom permission of Nelp case.
        """
        if request.method == "POST":
            org = request.data.get("org", None)
            return bool(
                roles.GlobalStaff().has_user(request.user) or roles.OrgInstructorRole(org=org).has_user(request.user)
            
        return super().has_permission(request, view)
```

johanv26: [653479a](https://github.com/eduNEXT/eox-nelp/pull/105/commits/653479a0360c80810ca70fecdfbb8140682de923)

andrey-canon (approved):

PR #106, opened 2023-11-10
Jlc/nid pipeline

johanv26:
## Description

[feat: add pipeline to match user with uid](https://github.com/eduNEXT/eox-nelp/commit/b2760b2a7105ac03c5a4a9854e0d6d80d5691206) 

This allow if the response send correct uid and there is a user
with username equal, this match that user with using that uid.

## Additional information
Jira story?
https://edunext.atlassian.net/browse/FUTUREX-621

## Checklist for Merge

- [ ] Updated documentation
- [ ] Rebased master/main
- [ ] Squashed commits

### DIFF
--- a/eox_nelp/third_party_auth/pipeline.py
+++ b/eox_nelp/third_party_auth/pipeline.py
@@ -8,6 +8,8 @@ from django.conf import settings
 from django.contrib.auth import logout
 from social_core.pipeline.social_auth import social_details as social_core_details
 
+from eox_nelp.third_party_auth.exceptions import EoxNelpAuthException
+
 
 def social_details(backend, details, response, *args, **kwargs):
     """This is an extension of `social_core.pipeline.social_auth.social_details` that allows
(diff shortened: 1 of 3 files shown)


eox_nelp/third_party_auth/pipeline.py
```diff
@@ -65,3 +75,36 @@ def close_mismatch_session(request, *args, user=None, **kwargs):  # pylint: disa
...
+    This is using the idp and the uid inspired in:
+    https://github.com/python-social-auth/social-core/blob/master/social_core/backends/saml.py#L296C23-L297
+    Raises:
+        EoxNelpAuthException: If someone tries to have staff or superuser permission using tpa.
+
+    Returns:
+        dict: Dict with user if matches, if not return None.
+    """
+    if user:
+        return None
```
andrey-canon, line 96: this should always return a dict, in this case that should be an empty dict since the pipeline [logic](https://github.com/python-social-auth/social-core/blob/master/social_core/backends/base.py#L119) will check the kind of instance

eox_nelp/third_party_auth/pipeline.py
```diff
@@ -65,3 +75,36 @@ def close_mismatch_session(request, *args, user=None, **kwargs):  # pylint: disa
...
+
+    idp = backend.get_idp(response["idp_name"])
+    uid = idp.get_user_permanent_id(response["attributes"])
+    user_match = backend.strategy.storage.user.get_user(username=uid)
+
+    if not user_match:
+        return None
```
andrey-canon, line 103: same comment as above

eox_nelp/third_party_auth/pipeline.py
```diff
@@ -9,6 +9,16 @@
 from social_core.pipeline.social_auth import social_details as social_core_details
 
 
+class EoxNelpAuthException(ValueError):
```
andrey-canon, line 12: could you move this to an exceptions file in this folder

eox_nelp/third_party_auth/tests/test_pipeline.py
```diff
@@ -0,0 +1,102 @@
...
+            - Strategy storage get_user method is called with desired params.
+        """
+        test_uid = "1222333444"
+        past_user, _ = User.objects.get_or_create(username=test_uid, is_superuser=True)
+        backend = Mock()
+        backend.get_idp.return_value.get_user_permanent_id.return_value = "1222333444"
+        backend.strategy.storage.user.get_user.return_value = past_user
+
+        self.assertRaises(EoxNelpAuthException, safer_associate_username_by_uid, backend, self.details, self.response)
+        backend.strategy.storage.user.get_user.assert_called_with(username=test_uid)
```
andrey-canon, line 82: this could be one test by using ddt

andrey-canon (changes requested):

eox_nelp/third_party_auth/pipeline.py
johanv26: I was thinking about this and why other pipes work using `None` in the return. 
https://github.com/eduNEXT/eox-tenant/blob/master/eox_tenant/pipeline.py#L25
https://github.com/python-social-auth/social-core/blob/master/social_core/pipeline/social_auth.py#L67

Then I found that there is an  [or-line](https://github.com/python-social-auth/social-core/blob/master/social_core/backends/base.py#L118
)  that allow use `None` changing it to `{}`.
image

PR #107, opened 2023-11-15
feat: add xapi actor filter

andrey-canon:
## Description

This allows to modify the actor field of event-routing-backends https://github.com/nelc/event-routing-backends/blob/and/add_openedx_filters/event_routing_backends/processors/xapi/transformer.py#L82

## Testing instructions
1. Use this version of event routing https://github.com/nelc/event-routing-backends/pull/5
2. add the settings        
```
       XAPI_AGENT_IFI_TYPE = "mbox"
       OPEN_EDX_FILTERS_CONFIG = {
            "event_routing_backends.processors.xapi.transformer.xapi_transformer.get_actor": {
                "pipeline": ["eox_nelp.openedx_filters.xapi.filters.XApiActorFilter"],
                "fail_silently": False,
            }
        }
```
3. Raise a valid event like, show_answer in a unit 
4.  Check the result, in my case I'm using aspects, so I can check that in the superset panel

### Before
image

### After
image

## Additional information

## Checklist for Merge

- [ ] Updated documentation
- [ ] Rebased master/main
- [ ] Squashed commits


eox_nelp/openedx_filters/tests/xapi/tests_filters.py
```diff
@@ -0,0 +1,70 @@
...
+        )
+        self.username = "xapi"
+        self.email = "xapi@example.com"
+        User.objects.update_or_create(username=self.username, email=self.email)
+
+    def test_user_does_not_exist(self):
+        """ Test case when the user is not found by the email.
+
+        Expected behavior:
+            - Returned value is the same as the given value.
```
johanv26, line 32: I thought that in this test would not have actor, but in this case is not modify

eox_nelp/openedx_filters/tests/xapi/tests_filters.py
```diff
@@ -0,0 +1,70 @@
...
+            - Returned value contains right mbox value.
+            - Returned value contains right name value.
+        """
+        actor = Agent(
+            mbox=self.email,
+        )
+
+        actor = self.filter.run_filter(result=actor)["result"]
+
+        self.assertEqual(f"mailto:{self.email}", actor.mbox)
```
johanv26, line 69: Who is adding the prefix `mailto` in the result.mbox? I checked that in the email is removed via the replace. But then we need to check its presence.

eox_nelp/openedx_filters/xapi/filters.py
```diff
@@ -0,0 +1,53 @@
...
+    XApiActorFilter: Modifies the standard ACtor in order to include the name attribute.
+"""
+from django.contrib.auth import get_user_model
+from openedx_filters import PipelineStep
+from tincan import Agent
+
+User = get_user_model()
+
+
+class XApiActorFilter(PipelineStep):
```
johanv26, line 13: This class could say something in the name  that we are trying to fit and actor for Nelp conditions...

johanv26 (commented):
Is curious that this code dont reload the server but works, I know that would run in the woker but is curious.Peek 2023-11-15 17-26

On Other hand is working, as supposed I think in my local env
Peek 2023-11-15 17-28
I only have little questions, but seems good and functional

eox_nelp/openedx_filters/xapi/filters.py
andrey-canon: That is not explicit but adding something like that could be redundant, I mean this is eox-nelp the plugin which extends the platform y the NELP way

eox_nelp/openedx_filters/tests/xapi/tests_filters.py
andrey-canon: tincan library https://github.com/RusticiSoftware/TinCanPython/blob/3.x/tincan/agent.py#L108

eox_nelp/openedx_filters/tests/xapi/tests_filters.py
andrey-canon: The method uses the given value to find a user, if the user is not found the method won't alter the result or will act as identity method

eox_nelp/openedx_filters/tests/xapi/tests_filters.py
johanv26: That make all clear for me,

johanv26 (approved):

PR #108, opened 2023-11-17
feat: add XApiBaseEnrollmentFilter

andrey-canon:
## Description

This filter allows to implement the required object structure by the NELC team https://edunext.atlassian.net/browse/FUTUREX-251

## Testing instructions

1. Use this version of event routing https://github.com/nelc/event-routing-backends/pull/5
2. add the settings        
```
        OPEN_EDX_FILTERS_CONFIG = {
            "event_routing_backends.processors.xapi.enrollment_events.base_enrollment.get_object": {
                "pipeline": ["eox_nelp.openedx_filters.xapi.filters.XApiBaseEnrollmentFilter"],
                "fail_silently": False,
            },
        }
```
3. This uses the language and the short_description studio fields, set them
4. Perform an enrollment or unenrollment event
5.  Check the result, in my case I'm using aspects, so I can check that in the superset panel

### Before
image

### After
image

## Additional information

## Checklist for Merge

- [ ] Updated documentation
- [ ] Rebased master/main
- [ ] Squashed commits

### DIFF
--- a/eox_nelp/openedx_filters/xapi/filters.py
+++ b/eox_nelp/openedx_filters/xapi/filters.py
@@ -2,12 +2,16 @@
 
 Filters:
     XApiActorFilter: Modifies the standard ACtor in order to include the name attribute.
+    XApiBaseEnrollmentFilter: Updates enrollment object definition.
 """
 from django.contrib.auth import get_user_model
 from openedx_filters import PipelineStep
-from tincan import Agent
+from tincan import Agent, LanguageMap
+
+from eox_nelp.utils import extract_course_id_from_string, get_course_from_id
 
 User = get_user_model()
+DEFAULT_LANGUAGE = "en"
 
 
 class XApiActorFilter(PipelineStep):
@@ -47,6 +51,50 @@ class XApiActorFilter(PipelineStep):
             )
         except User.DoesNotExist:
             pass
+
+        return {
+            "result": result
+        }
+
+
+class XApiBaseEnrollmentFilter(PipelineStep):
+    """This filter is designed to modify object attributes of an enrollment event, this will add
+    the description field and will change the name based on the course language.
+
+    How to set:
+        OPEN_EDX_FILTERS_CONFIG = {
+            "event_routing_backends.processors.xapi.enrollment_events.base_enrollment.get_object": {
+                "pipeline": ["eox_nelp.openedx_filters.xapi.filters.XApiBaseEnrollmentFilter"],
+                "fail_silently": False,
+            },
+        }
+    """
+
+    def run_filter(self, result):  # pylint: disable=arguments-differ
+        """Modifies name and description attributes of the activity's definition.
+
+        Arguments:
+            result <Activity>: Object activity for events related to enrollments.
+
+        Returns:
+            Activity: Modified activity.
+        """
+        course_id = extract_course_id_from_string(result.id)
+
+        if course_id:
+            # Get course course data
+            course = get_course_from_id(course_id)
+            display_name = course["display_name"]
+            description = course["short_description"]
+            course_language = course["language"] or DEFAULT_LANGUAGE  # Set default value if language is not found
+
+            # Create new attributes based on the course properties
+            definition_name = LanguageMap(**({course_language: display_name} if display_name is not None else {}))
+            description = LanguageMap(**({course_language: description} if description is not None else {}))
+
+            # Updates current result
+            result.definition.name = definition_name
+            result.definition.description = description
 
         return {
             "result": result
--- a/eox_nelp/utils.py
+++ b/eox_nelp/utils.py
@@ -2,7 +2,12 @@
 import re
 from copy import copy
 
+from opaque_keys.edx.keys import CourseKey
+
+from eox_nelp.edxapp_wrapper.course_overviews import get_course_overviews
+
 NATIONAL_ID_REGEX = r"^[1-2]\d{9}$"
+COURSE_ID_REGEX = r'(course-v1:[^/+]+(/|\+)[^/+]+(/|\+)[^/?]+)'
 
 
 def map_instance_attributes_to_dict(instance, attributes_mapping):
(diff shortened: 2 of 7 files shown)


eox_nelp/utils.py
```diff
@@ -2,7 +2,12 @@
 import re
 from copy import copy
 
+from opaque_keys.edx.keys import CourseKey
+
+from eox_nelp.edxapp_wrapper.course_overviews import CourseOverview
+
 NATIONAL_ID_REGEX = r"^[1-2]\d{9}$"
+COURSE_ID_REGEX = r'(course-v1:[^/+]+(/|\+)[^/+]+(/|\+)[^/?]+)'
```
johanv26, line 10: Where was extracted this Regex?
It seems a little different than 
https://github.com/openedx/edx-platform/blob/master/openedx/core/constants.py#L12
```python
>>> from django.conf import settings
>>> settings.COURSE_KEY_REGEX
'(?:[^/+]+(/|\\+)[^/+]+(/|\\+)[^/?]+)'
```

eox_nelp/utils.py
```diff
@@ -78,3 +83,39 @@ def is_valid_national_id(national_id, raise_exception=False):
...
+    Get Course object using the `course_id`.
+
+    Arguments:
+        course_id (str) :   ID of the course
+
+    Returns:
+        Course
+    """
+    course_key = CourseKey.from_string(course_id)
+    course_overviews = CourseOverview.get_from_ids([course_key])
```
johanv26, line 116: Why if the methods is called `get_course_from_id` (from one id), you use a method for multiple ids instead of the method for [one](https://github.com/openedx/edx-platform/blob/master/openedx/core/djangoapps/content/course_overviews/models.py#L388)??

eox_nelp/utils.py
```diff
@@ -78,3 +83,39 @@ def is_valid_national_id(national_id, raise_exception=False):
...
+
+    if course_overviews:
+        return course_overviews[0]
+
+    raise ValueError(f"Course with id {course_id} does not exist.")
```
johanv26, line 121: Why do you prefer the raise of an exception instead a `None` return as the CourseOverview do ?
https://github.com/openedx/edx-platform/blob/master/openedx/core/djangoapps/content/course_overviews/models.py#L415

eox_nelp/utils.py
```diff
@@ -78,3 +83,39 @@ def is_valid_national_id(national_id, raise_exception=False):
...
+        course_id (str) :   ID of the course
+
+    Returns:
+        Course
+    """
+    course_key = CourseKey.from_string(course_id)
+    course_overviews = CourseOverview.get_from_ids([course_key])
+
+    if course_overviews:
+        return course_overviews[0]
```
johanv26, line 119: Here I was testing and found like an error
Peek 2023-11-17 16-33

johanv26: Actually, this seem not return a list...
image

johanv26: https://github.com/openedx/edx-platform/blob/master/openedx/core/djangoapps/content/course_overviews/models.py#L438

johanv26 (changes requested):

eox_nelp/utils.py
andrey-canon: from that place the modification is the course-v1:

eox_nelp/utils.py
andrey-canon: This is a copy of [this](https://github.com/openedx/event-routing-backends/blob/master/event_routing_backends/helpers.py#L138)

eox_nelp/utils.py
andrey-canon: copy of https://github.com/openedx/event-routing-backends/blob/master/event_routing_backends/helpers.py#L138

eox_nelp/utils.py
andrey-canon: This method is a copy of [this](https://github.com/openedx/event-routing-backends/blob/master/event_routing_backends/helpers.py#L138) which implements  this line `course_overviews = get_course_overviews([course_key])` however I made a mistake  and I copied the logic from the method [get_course_overviews_from_ids](https://github.com/eduNEXT/edunext-platform/blob/ednx-release/mango.master.nelp/openedx/core/djangoapps/content/course_overviews/api.py#L61C5-L61C34) instead of the [get_course_overviews](https://github.com/eduNEXT/edunext-platform/blob/ednx-release/mango.master.nelp/openedx/core/djangoapps/content/course_overviews/api.py#L74C5-L74C25) method

andrey-canon: This method is a copy of [this](https://github.com/openedx/event-routing-backends/blob/master/event_routing_backends/helpers.py#L138) which implements  this line `course_overviews = get_course_overviews([course_key])` however I made a mistake  and I copied the logic from the method [get_course_overviews_from_ids](https://github.com/eduNEXT/edunext-platform/blob/ednx-release/mango.master.nelp/openedx/core/djangoapps/content/course_overviews/api.py#L61C5-L61C34) instead of the [get_course_overviews](https://github.com/eduNEXT/edunext-platform/blob/ednx-release/mango.master.nelp/openedx/core/djangoapps/content/course_overviews/api.py#L74C5-L74C25) method

andrey-canon: I copied the logic because we already have the CourseOverview model and was easier  however the other logic requires a backend, so it will be easier to implement new api backend

andrey-canon:
@johanv26 I have already applied some changes and tested that by making an enrollment in the about page, could please you try again :D

eox_nelp/openedx_filters/xapi/filters.py
```diff
@@ -51,3 +54,47 @@ def run_filter(self, result):  # pylint: disable=arguments-differ
...
+
+        if course_id:
+            # Get course course data
+            course = get_course_from_id(course_id)
+            display_name = course['display_name']
+            description = course['short_description']
+            course_language = course['language']
+
+            # Create new attributes based on the course properties
+            definition_name = LanguageMap(**({course_language: display_name} if display_name is not None else {}))
```
johanv26, line 91: Here you are validating `display_name` is not None, but also is possible that `course_language` is empty `None`
Peek 2023-11-20 13-18

johanv26 (commented):
I am testing again and pass last error, but after passing the course_overvieews using a list of ordered_dict, I had an error with `course_language` being None

eox_nelp/openedx_filters/xapi/filters.py
andrey-canon: @johanv26 How did you set the course language to None since that option is not available on studio ?

johanv26: Actually is a very good question. 
I tried to defined here, but is not working
image

But on the other hand my course_overview record of that object is empty.
From admin is difficult to change.
Peek 2023-11-20 16-41

Myabe there is a mismatch of mongo and mysql course_overview...

johanv26 (approved):
LGTM. Is working after some changes.
[feat: use default language when the course doesn't have that value](https://github.com/eduNEXT/eox-nelp/pull/108/commits/dacc260a82ca80ccaf24b9536c26563624af7356)
Peek 2023-11-20 16-46

andrey-canon:
Still I wonder how did you set to None the language, I couldn't do that I just considered that case

johanv26:
> Still I wonder how did you set to None the language, I couldn't do that I just considered that case, how did you create that course ?

@andrey-canon I was reseaching a little and I found that seems this backend use the mysql course_overview as source.

From studio is possible to set with `-` and that saves in mongo with `""`. By the way, the mysql course_overview record is not updated. But in mongo and what we see in studio yes.

Peek 2023-11-20 18-13


Here I also share you how in my local env I set arabic language, then using modulestore, yes my course language is `ar` but with the method of course_overview is `None`.
Peek 2023-11-20 18-21

PR #109, opened 2023-11-22
chore: update filters

andrey-canon:
## Description

This version updates the filters signature in order to support transformer arguments https://github.com/openedx/event-routing-backends/pull/374


johanv26 (approved):
LGTM, the only thing is that we have to ensure the image creation has the openedx filter version that brings the transformer arg.

PR #112, opened 2023-12-07
feat: add receiver which is able to emit an event the first time that…

andrey-canon:
## Description

 Add receiver which is able to emit an event the first time that a user completes problem.

https://edunext.atlassian.net/browse/FUTUREX-644

## Testing instructions
- create a course with some units
- complete one unit 
-  look for the `nelc.eox_nelp.initialized.course` in the logs

image

## Additional information

## Checklist for Merge

- [ ] Updated documentation
- [ ] Rebased master/main
- [ ] Squashed commits

### DIFF
--- a/eox_nelp/signals/receivers.py
+++ b/eox_nelp/signals/receivers.py
@@ -12,6 +12,7 @@ import logging
 
 from django.conf import settings
 from eox_core.edxapp_wrapper.grades import get_course_grade_factory
+from eventtracking import tracker
 from openedx_events.learning.data import CertificateData, CourseData, UserData, UserPersonalData
 
 from eox_nelp.notifications.tasks import create_course_notifications as create_course_notifications_task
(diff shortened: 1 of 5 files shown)


eox_nelp/signals/receivers.py
```diff
@@ -36,6 +37,29 @@ def block_completion_progress_publisher(instance, **kwargs):  # pylint: disable=
...
+    """This receiver is connected to the post_save BlockCompletion signal
+    and this checks how many BlockCompletion records exists, if there is only one record
+    the `nelc.eox_nelp.initialized.course` event will be emitted.
+
+    Args:
+        instance<Blockcompletion>: Instance of BlockCompletion model.
+    """
+    completion_blocks = instance.user_learning_context_completion_queryset(instance.user, instance.context_key)
+
+    if len(completion_blocks) == 1:
```
johanv26, line 50: There is only for initalized and not for finish course. 
That the reason the qs is only 1 when there is only one in the completion queryset per user and course.?
image

johanv26 (approved):
I tested in my local env and seems working.

Screenshot from 2023-12-11 17-58-34

eox_nelp/signals/receivers.py
andrey-canon: > There is only for initalized and not for finish course event? That the reason the qs is only 1 when there is only one in the completion queryset per user and course.? image

The finish event is the course completion event that is generated by aggregators  and I didn't catch your second question

PR #120, opened 2023-12-29
refactor!: this modifies the enrollment filter in order to generalize…

andrey-canon:
## Description

This includes the following:

1. Rename the `XApiBaseEnrollmentFilter` class to  `XApiCourseObjectFilter` so now that filter is able to modify any event with a course as object
2. Update course language logic for  `XApiCourseObjectFilter`, so now if the language is `en` that will be replaced to `en-US`

## Testing instructions

1. Set the right setting, the doc string was update with a new version that includes more filters
2. Generate the event, you could complete a course or makes a enrollment or complete a component 
3. check the result

image

image

## Additional information

Include anything else that will help reviewers and consumers understand the change.
- Does this change depend on other changes elsewhere?
- Any special concerns or limitations? For example: deprecations, migrations, security, or accessibility.
- Link to other information about the change, such as Jira issues, GitHub issues, or Discourse discussions.

## Checklist for Merge

- [ ] Tested in a remote environment
- [ ] Updated documentation
- [ ] Rebased master/main
- [ ] Squashed commits

PR #121, opened 2023-12-29
refactor!: generalize XApiBaseProblemsFilter to support module and qu…

andrey-canon:
## Description

This is a refactor in order to use the same filter for objects with definition type equals to module or question

## Testing instructions

1. Set the right setting, the doc string was update with a new version that includes more filters
2. Generate the event, you could attempt any problem
3. check the result

image

## Additional information

## Checklist for Merge

- [ ] Updated documentation
- [ ] Rebased master/main
- [ ] Squashed commits

### DIFF
--- a/eox_nelp/openedx_filters/tests/xapi/tests_filters.py
+++ b/eox_nelp/openedx_filters/tests/xapi/tests_filters.py
@@ -3,10 +3,10 @@
 Classes:
     XApiActorFilterTestCase: Tests cases for XApiActorFilter filter class.
     XApiCourseObjectFilterTestCase: Test cases for XApiCourseObjectFilter filter class.
-    XApiBaseProblemsFilterTestCase: Test cases for XApiBaseProblemsFilter filter class.
+    XApiModuleQuestionObjectFilterTestCase: Test cases for XApiModuleQuestionObjectFilter filter class.
     XApiVerbFilterTestCase: Test cases for XApiVerbFilter filter class.
 """
-from ddt import data, ddt
+from ddt import data, ddt, unpack
 from django.contrib.auth import get_user_model
 from django.test import TestCase
 from mock import Mock, patch
@@ -16,8 +16,8 @@ from eox_nelp.edxapp_wrapper.event_routing_backends import constants
 from eox_nelp.edxapp_wrapper.modulestore import modulestore
 from eox_nelp.openedx_filters.xapi.filters import (
     XApiActorFilter,
-    XApiBaseProblemsFilter,
     XApiCourseObjectFilter,
+    XApiModuleQuestionObjectFilter,
     XApiVerbFilter,
 )
 from eox_nelp.processors.xapi.constants import DEFAULT_LANGUAGE
@@ -189,8 +189,8 @@ class XApiCourseObjectFilterTestCase(TestCase):
 
 
 @ddt
-class XApiBaseProblemsFilterTestCase(TestCase):
-    """Test class for XApiBaseProblemsFilterr filter class."""
+class XApiModuleQuestionObjectFilterTestCase(TestCase):
+    """Test class for XApiModuleQuestionObjectFilterr filter class."""
 
     def setUp(self):
         """Setup common conditions for every test case"""
@@ -199,22 +199,16 @@ class XApiBaseProblemsFilterTestCase(TestCase):
             "data.problem_id": "block-v1:edx+CS105+2023-T3+type@problem+block@0221040b086c4618b6b2b2a554558",
             "course_id": "course-v1:edx+CS105+2023-T3",
         }
-        self.filter = XApiBaseProblemsFilter(
+        self.filter = XApiModuleQuestionObjectFilter(
             filter_type="event_routing_backends.processors.xapi.problem_interaction_events.base_problems.get_object",
-            running_pipeline=["eox_nelp.openedx_filters.xapi.filters.XApiBaseProblemsFilter"],
+            running_pipeline=["eox_nelp.openedx_filters.xapi.filters.XApiModuleQuestionObjectFilter"],
         )
         self.transformer = Mock()
         self.transformer.event = {"name": "edx.grades.problem.submitted"}
         self.transformer.get_data.side_effect = lambda x: self.default_values[x]
         self.transformer._get_submission.return_value = {}  # pylint: disable=protected-access
         self.course = {"language": "ar"}
-        self.activity = Activity(
-            id="https://example.com/xblock/block-v1:edx+CS105+2023-T3+type@problem+block@0221040b086c4618b6b2b2a554558",
-            definition=ActivityDefinition(
-                type="http://adlnet.gov/expapi/activities/question",
-                name=LanguageMap(en="old-testing-course"),
-            ),
-        )
+        self.activity = None
 
         item = Mock()
         item.markdown = ""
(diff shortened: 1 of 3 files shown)


eox_nelp/openedx_filters/xapi/filters.py
```diff
@@ -126,32 +130,32 @@ def run_filter(self, transformer, result):  # pylint: disable=arguments-differ
...
+            item = modulestore().get_item(usage_key)
+            label = get_item_label(item)
 
-        # Get label from module descriptor block.
-        usage_key = UsageKey.from_string(transformer.get_data("data.problem_id"))
-        item = modulestore().get_item(usage_key)
-        label = get_item_label(item)
+            # Get course languge block.
+            course = get_course_from_id(transformer.get_data('course_id'))
+            course_language = course.get("language")
```
johanseto, line 149: I think that we could create a function for language if this piece of code is multiple used.
https://github.com/eduNEXT/eox-nelp/pull/120/files#diff-7f85cc618f32f4937416b9a541239a453597fac05c0552e3c34011f6faa0aaadR100-R104

eox_nelp/openedx_filters/xapi/filters.py
```diff
@@ -107,17 +107,21 @@ def run_filter(self, transformer, result):  # pylint: disable=arguments-differ,
...
-    the name based on the course language.
+class XApiModuleQuestionObjectFilter(PipelineStep):
+    """This filter is designed to modify object attributes of events whose object type is a module
+    or a question, this will add the description field and will change the name based on the course language.
 
     How to set:
         OPEN_EDX_FILTERS_CONFIG = {
             "event_routing_backends.processors.xapi.problem_interaction_events.base_problems.get_object": {
-                "pipeline": ["eox_nelp.openedx_filters.xapi.filters.XApiBaseProblemsFilter"],
+                "pipeline": ["eox_nelp.openedx_filters.xapi.filters.XApiModuleQuestionObjectFilter"],
```
johanseto, line 117: I like more the name related `problems`, why you change it to `questions`?

eox_nelp/openedx_filters/xapi/filters.py
```diff
@@ -126,32 +130,32 @@ def run_filter(self, transformer, result):  # pylint: disable=arguments-differ
 
         Arguments:
             transformer <XApiTransformer>: Transformer instance.
-            result <Activity>: Object activity for events related to enrollments.
+            result <Activity>: Module or Problem object activity for any kind of event.
 
         Returns:
             Activity: Modified activity.
         """
-        # Following line is hard-coded since this logic has not been tested with other events.
```
johanseto, line 134: Why you remove the exit of events different of "edx.grades.problem.submitted"

eox_nelp/openedx_filters/tests/xapi/tests_filters.py
```diff
@@ -263,41 +278,49 @@ def test_update_description(self, get_course_mock):
...
             - Definition name has the default key language.
         """
         get_course_mock.return_value = {}
+        self.set_activity(definition_type)
 
         returned_activity = self.filter.run_filter(transformer=self.transformer, result=self.activity)["result"]
 
         self.assertEqual([DEFAULT_LANGUAGE], list(returned_activity.definition.name.keys()))
 
     def test_invalid_event(self):
```
johanseto, line 311: For this test I have some doubts about why if in the pipe you remove the normal exist of the event. Why this test seem without problem?.

eox_nelp/openedx_filters/tests/xapi/tests_filters.py
```diff
@@ -205,16 +199,33 @@ def tearDown(self):
...
+            )
+        )
+
+    @data(
+        (constants.XAPI_ACTIVITY_MODULE, None),
+        (constants.XAPI_ACTIVITY_MODULE, ""),
+        (constants.XAPI_ACTIVITY_QUESTION, None),
+        (constants.XAPI_ACTIVITY_QUESTION, ""),
+    )
+    @unpack
```
johanseto, line 218: Why did you choose to use the unpack instead of using an object like a dict. And inside the test use the dict to get the keys related. eg `definiton_type`?

eox_nelp/openedx_filters/tests/xapi/tests_filters.py
```diff
@@ -263,41 +278,49 @@ def test_update_description(self, get_course_mock):
...
         """
         self.transformer.event = {"name": "other-event"}
         activity = Activity(
             id="empty-activity",
+            definition=ActivityDefinition(
+                type="invalid-type",
+            )
         )
 
         returned_activity = self.filter.run_filter(transformer=self.transformer, result=activity)["result"]
```
johanseto, line 326: Like in the above comment, due the removal of this line https://github.com/eduNEXT/eox-nelp/blob/af74b95fba3c903505f78065cb873fe8240a3616/eox_nelp/openedx_filters/xapi/filters.py#L135
In the past test for eg modulestore were not called. But in this test yes, https://github.com/eduNEXT/eox-nelp/pull/121/files#diff-7f85cc618f32f4937416b9a541239a453597fac05c0552e3c34011f6faa0aaadR144

So I don't understand the invalid type if all the logic is also run. Maybe the logic doesn't change anything 🤔???🤔🤔
Or you are only filtering by the definiton and not the event-name.
https://github.com/eduNEXT/eox-nelp/blob/de1c32e0b8eba3447cc306c11e5dd65cc7812de4/eox_nelp/openedx_filters/xapi/filters.py#L138

eox_nelp/openedx_filters/tests/xapi/tests_filters.py
andrey-canon: As you have described that implementation requires more changes and more code

eox_nelp/openedx_filters/xapi/filters.py
andrey-canon: As I mentioned in the doc-string this filter works with events that have a module or question object type, so this is not a unique filter for problems, and why question because a  `edx.grades.problem.submitted` event has a question definition type https://github.com/nelc/event-routing-backends/blob/master/event_routing_backends/processors/xapi/event_transformers/problem_interaction_events.py#L40

eox_nelp/openedx_filters/xapi/filters.py
andrey-canon: Because now this use the definition type

eox_nelp/openedx_filters/tests/xapi/tests_filters.py
andrey-canon: I think this is clearer with the other answers, if not let me know

eox_nelp/openedx_filters/tests/xapi/tests_filters.py
andrey-canon: the criteria of the event changed this has an invalid definition type that's why this works, to be honest and not sure if I understood your question

eox_nelp/openedx_filters/xapi/filters.py
andrey-canon: IMO this change makes more sense in a different PR, [here](https://github.com/eduNEXT/eox-nelp/pull/120) you mentioned the necessity of using EN_US, so I think this change should be included in that pr

johanseto (approved):
LGTM. With the explanation and this line I understood.
https://github.com/eduNEXT/eox-nelp/pull/121/files#r1439654134

PR #122, opened 2024-01-03
fix: get display name from module descriptor instead of tracker context

andrey-canon:
## Description

This pr considers two things

1. the line `transformer.get_data('display_name')` obtains the value from the tracker [context](https://github.com/openedx/event-tracking/blob/706ac528d86962d5d4dedda992f609a06db042ff/eventtracking/tracker.py#L41)  however that context is not updated for the completion-aggregator library when the event is emitted  so as you can see in the following images  the context is the same for problems, verticals, sequentials and chapters 

image

image

image

After these changes the result is the following

image

image
image

2. Even if the context were updated that context just works on synchronous processes so this implementation contemplates async tasks

## Additional information

## Checklist for Merge

- [ ] Updated documentation
- [ ] Rebased master/main
- [ ] Squashed commits

### DIFF
--- a/eox_nelp/openedx_filters/tests/xapi/tests_filters.py
+++ b/eox_nelp/openedx_filters/tests/xapi/tests_filters.py
@@ -195,7 +195,6 @@ class XApiModuleQuestionObjectFilterTestCase(TestCase):
     def setUp(self):
         """Setup common conditions for every test case"""
         self.default_values = {
-            "display_name": "testing-course",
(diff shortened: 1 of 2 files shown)


johanseto (approved):
I understood the fix. Actually I don`t know why the transformer wants to be unique? and "return duplicate of duplicate of duplicate". Nice catch that for the unit,sequential and vertical the transformer returns the same.
Now getting the `display_name` from modulestore is always related and its the proper of the block

PR #124, opened 2024-01-05
feat: add context filter

andrey-canon:
## Description

This adds extra fields based on the settings to the context key of any statement 
## Testing
1 include the extra settings
```
    "XAPI_EXTRA_CONTEXT": {
        "extensions": {
            "https://nelc.gov.sa/extensions/platform": {
                "name": {
                    "ar-SA": "\u0627\u0644\u062a\u0642\u0646\u064a\u0627\u062a \u0627\u0644\u0645\u062a\u0642\u062f\u0645\u0629 \u0644\u0644\u062a\u062f\u0631\u064a\u0628",
                    "en-US": "Leading Tech for Training"
                }
            }
        },
        "instructor": {
            "mbox": "mailto:ik@example.com",
            "name": " Ibrahim Khalid "
        },
        "language": "ar-SA",
        "platform": "SREI"
    },
```
2. raised any event
3. check the context filed

### Before
image

### After
image

## Additional information

## Checklist for Merge

- [ ] Updated documentation
- [ ] Rebased master/main
- [ ] Squashed commits

### DIFF
--- a/eox_nelp/openedx_filters/xapi/filters.py
+++ b/eox_nelp/openedx_filters/xapi/filters.py
@@ -5,10 +5,11 @@ Filters:
     XApiCourseObjectFilter: Updates course object definition.
     XApiVerbFilter: Updates verba display language key.
 """
+from django.conf import settings
 from django.contrib.auth import get_user_model
-from opaque_keys.edx.keys import UsageKey
+from opaque_keys.edx.keys import CourseKey, UsageKey
 from openedx_filters import PipelineStep
-from tincan import Agent, LanguageMap
+from tincan import Agent, Group, LanguageMap
 
 from eox_nelp.edxapp_wrapper.event_routing_backends import constants
 from eox_nelp.edxapp_wrapper.modulestore import modulestore
@@ -152,7 +153,7 @@ class XApiModuleQuestionObjectFilter(PipelineStep):
             display_name = item.display_name
             label = get_item_label(item)
 
-            # Get course languge block.
+            # Get course language block.
             course = get_course_from_id(transformer.get_data('course_id'))
             course_language = course.get("language")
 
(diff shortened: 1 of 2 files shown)


eox_nelp/openedx_filters/xapi/filters.py
```diff
@@ -198,3 +199,45 @@ def run_filter(self, transformer, result):  # pylint: disable=arguments-differ,
...
+            },
+        }
+    """
+
+    def run_filter(self, transformer, result):  # pylint: disable=arguments-differ, unused-argument
+        """Modifies statement context for any event
+
+        Arguments:
+            transformer <XApiTransformer>: Transformer instance.
+            result <Context>: Context related to an event.
```
johanseto, line 221: Here could be an example of the shape of the XAPI_EXTRA_CONTEXT. I mean the key is the attribute name and the value could be a dict with definitions.

eox_nelp/openedx_filters/xapi/filters.py
```diff
@@ -198,3 +199,45 @@ def run_filter(self, transformer, result):  # pylint: disable=arguments-differ,
...
+        for key, value in extra_context.items():
+            if not hasattr(result, key):
+                break
+
+            attribute = getattr(result, key)
+
+            if isinstance(attribute, dict):
+                attribute.update(value)
+            else:
+                attribute = value
```
johanseto, line 237: If the attribute is not a dict it would override all the attribute so that setting has more priority. 
The filter would have a name like `SET`. Is like the operation you are realizing, but I am thinking it because if is dict it could preserve some other values of the dict...

eox_nelp/openedx_filters/xapi/filters.py
```diff
@@ -198,3 +199,45 @@ def run_filter(self, transformer, result):  # pylint: disable=arguments-differ,
...
+            transformer <XApiTransformer>: Transformer instance.
+            result <Context>: Context related to an event.
+
+        Returns:
+            Context: Modified context.
+        """
+        extra_context = getattr(settings, "XAPI_EXTRA_CONTEXT", {})
+
+        for key, value in extra_context.items():
+            if not hasattr(result, key):
```
johanseto, line 229: Is not possible to add a new attribute if the result object doesn't have the attribute.
So this filter only updates attributes that the result has.
This could be specified in the docstring, but maybe we could need to add extra attributes or definitions hard-coded.

johanseto: There is a text with that. But why you wanted it in that way?
https://github.com/eduNEXT/eox-nelp/blob/39a9cc86fbc25f29a015fb2e1ea06baa8316330a/eox_nelp/openedx_filters/tests/xapi/tests_filters.py#L456-L468

andrey-canon: Tincan library just supports specific fields, so this was implemented in this way to follow the tincan standard fields https://github.com/RusticiSoftware/TinCanPython/blob/3.x/tincan/context.py#L42

eox_nelp/openedx_filters/xapi/filters.py
andrey-canon: the run_filter name can not be changed  since openedx filter depends on that or if you refer to the class name this is just following the other filter name convention `XApiSomethingFilter ` so let me ask you for you what is an **openedx** filter ?

johanseto: I meant the class, but I am ok with the convention.

andrey-canon: keep in mind that is not a standard that is a convention that I adopted  since I think the `Filter` suffix makes reference to set, update or edit actions in the openedx-filter context, IMO filter was not the best name for that feature but it's what we have

johanseto (approved):
I am going to approve. The only thing missing was the shape example for `XAPI_EXTRA_CONTEXT` setting in some dosctring inside 
`XApiContextFilter` class

PR #127, opened 2024-01-26
Jlc/palm mig/language middleware

johanseto:
## Description

Migrate the language per cookie usage from edx-platform to plugin.

This was inspired in the `COOKIES` manipulation by django.https://github.com/django/django/blob/main/django/core/handlers/wsgi.py#L100-L102
## Testing instructions
1. Checkout you edx-platform to `open-release/palm.nelp`.

2. Install eox-nelp using this branch or using editable way.
eg
```
pip install git+https://github.com/eduNEXT/eox-nelp.git@jlc/palm-mig/language-middleware --force-reinstall --no-deps
```
3. add the middleware in setting using the following config.
``` python
index_nelp_lang = index_nelp_lang = MIDDLEWARE.index("openedx.core.djangoapps.lang_pref.middleware.LanguagePreferenceMiddleware") + 1
MIDDLEWARE.insert(index_nelp_lang, "eox_nelp.middleware.PreserveUserLanguageCookieMiddleware")

```
[Screencast from 26-01-24 17:01:30.webm](https://github.com/eduNEXT/eox-nelp/assets/51926076/0db6895c-38a4-469f-9dd4-7a8937acea30)
### Before
[Screencast from 26-01-24 16:54:14.webm](https://github.com/eduNEXT/eox-nelp/assets/51926076/4feb48ca-8eb9-4b2b-9760-2a032652ee25)

### After
[Screencast from 26-01-24 17:02:47.webm](https://github.com/eduNEXT/eox-nelp/assets/51926076/1b1cacff-0bef-4ee1-b58d-0ad6f99c2421)
## Additional information

**PRS related**
original-pr with discussion: https://github.com/nelc/edx-platform/pull/5
migration pr: https://github.com/eduNEXT/edunext-platform/pull/672
## Checklist for Merge

- [ ] Updated documentation
- [ ] Rebased master/main
- [ ] Squashed commits

### DIFF
--- a/eox_nelp/middleware.py
+++ b/eox_nelp/middleware.py
@@ -5,6 +5,9 @@ Required NELP middlewares that allow to customize edx-platform.
 classes:
     ExtendedProfileFieldsMiddleware: Set extended_profile_fields in registration form.
 """
+from django.conf import settings
+from django.http import parse_cookie
+from django.utils.deprecation import MiddlewareMixin
 from django.utils.translation import gettext_lazy as _
 
 from eox_nelp.edxapp_wrapper.site_configuration import configuration_helpers
@@ -133,3 +136,21 @@ class ExtendedProfileFieldsMiddleware:
             # pylint: disable=protected-access
             form_instance._add_field_with_configurable_select_options(field, label, form_desc, required=required)
         return handler
+
+
+class PreserveUserLanguageCookieMiddleware(MiddlewareMixin):
+    """This middleware ensure that in the COOKIES property the LANGUAGE_COOKIE_NAME
+    key has to be the cookie sent by the user and not other,
+    because it could be modified previously by other.
+    eg.
+    https://github.com/openedx/edx-platform/blob/open-release/palm.master/openedx/
+    core/djangoapps/lang_pref/middleware.py#L61-L62
+    """
+    def process_request(self, request):
+        """Process the request to change the language cookie based in original cookie value."""
+        original_user_language_cookie = parse_cookie(request.META.get("HTTP_COOKIE", "")).get(
+            settings.LANGUAGE_COOKIE_NAME
+        )
+
+        if original_user_language_cookie:
+            request.COOKIES[settings.LANGUAGE_COOKIE_NAME] = original_user_language_cookie
--- a/eox_nelp/middleware.py
+++ b/eox_nelp/middleware.py
@@ -7,7 +7,6 @@ classes:
 """
 from django.conf import settings
 from django.http import parse_cookie
-from django.utils.deprecation import MiddlewareMixin
 from django.utils.translation import gettext_lazy as _
 
 from eox_nelp.edxapp_wrapper.site_configuration import configuration_helpers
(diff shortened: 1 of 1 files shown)


eox_nelp/middleware.py
```diff
@@ -6,7 +6,8 @@
     ExtendedProfileFieldsMiddleware: Set extended_profile_fields in registration form.
 """
 from django.conf import settings
-from django.http import Http404
+from django.http import Http404, parse_cookie
+from django.utils.deprecation import MiddlewareMixin
```
andrey-canon, line 10: why are you using deprecation ?

johanseto: Because openedx is using it?
https://github.com/openedx/edx-platform/blob/open-release/palm.master/openedx/core/djangoapps/lang_pref/middleware.py#L5
https://github.com/openedx/edx-platform/blob/open-release/palm.master/openedx/core/djangoapps/dark_lang/middleware.py#L15

andrey-canon: IMO that's is not a valid reason, probably that is a old middleware that openedx had to update when they made the django upgrade, however this is a new middleware that you're making from scratch

johanseto: @andrey-canon Actually is somehting confusinf in the nomenclature of django.
https://stackoverflow.com/questions/52913333/why-middleware-mixin-declared-in-django-utils-deprecation-py
Is like a help to create new middleware or to support old styles middleware.
https://stackoverflow.com/questions/52913333/why-middleware-mixin-declared-in-django-utils-deprecation-py

Some base middleware of django are inheriting from it.

andrey-canon: image


> Some base middleware of django are inheriting from it.

This is the same argument that you exposed in the first comment but with django, my answer is the same those are old middlewares that they made compatible with the new middleware structure 
my question would be do you have any reason to use the structure that django deprecated in the past, why do you think the deprecated version is better than the "new" django way ?

johanseto: No, but I think that Django makes this middleware-mixin with the idea of being compatible with both ways: MIDDLEWARE AND MIDDLEWARE_CLASSES. So I prefer the compatibility of both ways for the middleware. 
image

andrey-canon: MIDDLEWARE_CLASSES was deprecated in Django  1.10 and probably you won't find any use of that in later versions

johanseto: d47f188fb210626fb6b2baac46c1f84b09d6b408

eox_nelp/middleware.py
```diff
@@ -5,6 +5,8 @@
 classes:
     ExtendedProfileFieldsMiddleware: Set extended_profile_fields in registration form.
```
andrey-canon, line 6: could you please add the new middleware here

johanseto: [dbfda6b](https://github.com/eduNEXT/eox-nelp/pull/127/commits/dbfda6b358edbc876e43c300976cf5585023b8f8)

andrey-canon (approved):

PR #130, opened 2024-02-05
Jlc/palm mig/patch generate passwords

johanseto:
## Description
This PR is a monkey patch to generate_password.
This was a PR of edx-platform but now this is included in eox-nelp.

As the feature of the [PR](https://github.com/eduNEXT/edunext-platform/pull/676)  needs django-password-validators this PR add to eox-nelp the possibility to install it.

Also edx-django-utils is included in requirements, but is a normal package of edx-platform.

I added a logger to know that the password is generated via eox-nelp.

## Testing instructions
- install this package with this branch

- Configure auth password validators in settings with custom rules, you could use [this](https://github.com/nelc/manifest-sagar/blob/8b950ae70f66a0232a6f8336c3f6dfb421fffd74/environments/prod/openedx/env/base/apps/openedx/settings/lms/production.py#L332) as eg.
``` python
AUTH_PASSWORD_VALIDATORS = [
...
]
```

- add django-password-validators in installed apps.
```python
INSTALLED_APPS += [
    "django_password_validators",
    "django_password_validators.password_history",
]
```
- Check into `/register` in the password field that you custom rules are loaded.
image

- Use a saml integration in your local env and complete a register flow, then check that the process is complete without problem.  Check into logs that eox-nelp geneate the password.
image

## After
[Screencast from 05-02-24 15:44:04.webm](https://github.com/eduNEXT/eox-nelp/assets/51926076/fd4a9bb3-09b9-4e98-bee7-b1ace3e2fc38)

## Additional information
PRs related
https://github.com/nelc/edx-platform/pull/4
https://github.com/eduNEXT/edunext-platform/pull/676

### DIFF
--- a/eox_nelp/init_pipeline.py
+++ b/eox_nelp/init_pipeline.py
@@ -20,6 +20,7 @@ def run_init_pipeline():
     set_mako_templates()
     register_xapi_transformers()
     update_permissions()
+    patch_generate_password()
 
 
 def patch_user_gender_choices():
@@ -82,3 +83,14 @@ def update_permissions():
         | rules.HasAccessRule("staff")
         | rules.HasAccessRule("instructor")
     )
+
+
+def patch_generate_password():
+    """This method patch `generate_password` of edx_django_util package,
+    with custom nelp `generate_password`.
+    """
+    # pylint: disable=import-outside-toplevel, unused-import
+    from edx_django_utils import user
+
+    from eox_nelp.user_authn.utils import generate_password
+    user.generate_password = generate_password
--- a/eox_nelp/user_authn/utils.py
+++ b/eox_nelp/user_authn/utils.py
@@ -1,0 +1,94 @@
+import logging
+import random
+import string
+
+from django.conf import settings
+
+LOGGER = logging.getLogger(__name__)
+
+
+def password_rules():
+    """
+    Inspect the validators defined in AUTH_PASSWORD_VALIDATORS and define
+    a rule list with the set of available characters and their minimum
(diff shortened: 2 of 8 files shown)


andrey-canon (approved):
LGTM, please rebase, add this feature to the testing list  and verify its behavior in alpha

eox_nelp/init_pipeline.py
```diff
@@ -82,3 +83,14 @@ def update_permissions():
...
+
+def patch_generate_password():
+    """This method patch `generate_password` of edx_django_util package,
+    with custom nelp `generate_password`.
+    """
+    # pylint: disable=import-outside-toplevel, unused-import
+    from edx_django_utils import user
+
+    from eox_nelp.user_authn.utils import generate_password
+    user.generate_password = generate_password
```
OmarIthawi, line 96: Thanks @johanseto for the fix today.

I believe we should patch the generate password locally inside the importing modules:

```
from lms.djangoapps.support.views import manage_user
manage_user.generate_password = generate_password
```

Otherwise the imports won't be updated in `manage_user` due to the patch being too late in the import order.

What do you think?

OmarIthawi: Forking edx_django_utils is also an option.

johanseto: Thanks @OmarIthawi  the patching locally is managed this PR.
https://github.com/eduNEXT/eox-nelp/pull/208/files
Is doing what you recommend using eox-tenant and seems working. Let us check out the behavior.

PR #140, opened 2024-03-14
feat: implement multiple configurations

andrey-canon:
## Description
This implements the [multiple configurations](https://webpack.js.org/configuration/configuration-types/#exporting-multiple-configurations)  webpack structure to  allow plugin modifications per plugin

## Testing instructions

1. Run npm run build
2. Go to the stats and feedback page and verify
3. Run npm run dev
4.  Go to the stats and feedback page and verify

## Checklist for Merge

- [ ] Updated documentation
- [ ] Rebased master/main
- [ ] Squashed commits


johanseto (approved):
LGTM, I didnt now that is possible to use [multiple-configurations](https://webpack.js.org/configuration/configuration-types/#exporting-multiple-configurations) in webpack.
I know this is the only way that seems work for us, but by curiosity: 
How did you decide to move from [multiple-entries](https://webpack.js.org/concepts/entry-points/#multi-page-application) to multiple-configs?


Also is working in my local env:
Peek 2024-03-15 10-00

webpack.dev.config.js
```diff
@@ -1,7 +1,15 @@
 const { merge } = require('webpack-merge');
 const commonConfig = require('./webpack.common.config');
 
-module.exports = merge(commonConfig, {
-  mode: 'development',
-  devtool: 'eval-source-map',
+const configuration = [];
```
johanseto, line 4: Which are the cases where we use `npm run dev` xD?

andrey-canon: `npm run build` generates the minify version of the files with the `npm run dev`  you can generate files that are easier to read and if I'm not wrong that generates the whole file tree in the browser console instead of just one js file, hot reloading but I think that's not working

PR #147, opened 2024-04-09
feat: return signup source records in learner metrics

andrey-canon:
## Description

This adds users from user singup source records to the get_learners_metric result, basically this metric will take into account two conditions  user that are enrolled in a tenant course and users that completed the signup process at that tenant.

## Testing instructions
1. Go to `/eox-nelp/api/stats/v1/tenant/`
2. Check the value of the field learners
3. Verify that the returned value is  (Total of  enrollments for course of the current tenant ) + (Total of  signup source records that belong to the current tenant)

**Note**

The metric doesn't return duplicates, so if the user completed the signup process in the current tenant and has at least one enrollment, the metric will return  1

On development environments the signupsource records uses the port, probably you will have to remove that, since the request.site.domain doesn't use the port

### DIFF
--- a/eox_nelp/edxapp_wrapper/test_backends/users_p_v1.py
+++ b/eox_nelp/edxapp_wrapper/test_backends/users_p_v1.py
@@ -1,0 +1,21 @@
+"""Test backend for eox-coreusers module."""
+from django.contrib.auth.models import User
+from django.db import models
+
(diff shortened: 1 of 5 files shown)


johanseto (commented):
LGTM. I think this is only adding to the count users that registered in the tenant but didn't enroll a course.

PR #150, opened 2024-05-15
refactor: implement authentication class and authenticator

andrey-canon:
## Description
This is a refactor of the api client module, instead of implementing a REST api client with an specific authentication method this creates a general Rest api client that allows to exchange the authentication method.

## Testing instructions
One option is to follow the instruction of this [file](https://docs.google.com/document/d/1uzGcWyYOBqbDxzUNgtsZSk9e31P6urgzXjmM-AocZCk/edit) testing case 2 and  28 however some testing services are not available so it is not possible to test the whole integration 

Other option is to compare the session that is generated with the master branch  and the session that is generated with this branch, example
image


eox_nelp/api_clients/__init__.py
```diff
@@ -6,19 +6,17 @@
...
+from eox_nelp.api_clients.authenticators import UnAuthenticatedAuthenticator
 
 LOGGER = logging.getLogger(__name__)
 
 
 class AbstractApiClient(ABC):
-    """Abstract api client class, this implement a basic authentication method and defines methods POST and GET"""
+    """Abstract api client class, this defines common API client methods."""
 
+    authentication_class = UnAuthenticatedAuthenticator
```
johanseto, line 19: There is a reason to use this one as default?

johanseto (approved):
I like the refactor to use authenticator classes to the API client. Better to polyform the authenticators attribute instead of all API Clients.

As reminder, rename the PR xD

PR #152, opened 2024-05-16
feat: add pearson_vue module and rti backend

andrey-canon:
## Description

This add a class that executes pipeline steps and a general task that implements the mentioned class

## Testing instructions

1. Create a testing method
```
def test(pipeline_index, **kwargs):
    print(pipeline_index)
```
2. include the test method in the pipeline list
```
    def get_pipeline(self):
        """
        Returns the RTI pipeline, which is a list of functions to be executed.
        """
        return [
            test,
            test,
            test,
            test,
            test,
            test,
            test,
        ]
```
3. Go to the lms shell `./manage.py lms shell`
4. import the task `from eox_nelp.pearson_vue.tasks import real_time_import_task`
5. call the task with some parameters `real_time_import_task.apply(user_id=1, course_id="hello")`
image

## Additional information

## Checklist for Merge

- [ ] Updated documentation
- [ ] Rebased master/main
- [ ] Squashed commits

### DIFF
--- a/eox_nelp/pearson_vue/rti_backend.py
+++ b/eox_nelp/pearson_vue/rti_backend.py
@@ -1,0 +1,49 @@
+"""
+This module provides the RealTimeImport class, which is responsible for orchestrating the RTI pipeline
+and executing various processes related to rti.
+
+Classes:
+    RealTimeImport: Class for managing RTI operations and executing the pipeline.
+"""
+
+
+class RealTimeImport:
+    """
+    Class for managing RTI (Real Time Import) operations and executing the pipeline.
+
+    Attributes:
+        backend_data (dict): A dictionary containing backend-specific data.
+
+    Methods:
+        run_pipeline(): Executes the RTI pipeline by iterating through the pipeline functions.
+        get_pipeline(): Returns the RTI pipeline, which is a list of functions to be executed.
+    """
+
+    def __init__(self, **kwargs):
+        """
+        Initializes the RealTimeImport instance with the provided keyword arguments.
+
+        Args:
+            **kwargs: Additional keyword arguments to configure the RealTimeImport instance.
+        """
+        self.backend_data = kwargs.copy()
+
+    def run_pipeline(self):
+        """
+        Executes the RTI pipeline by iterating through the pipeline functions.
+        """
+        pipeline = self.get_pipeline()
+        pipeline_index = self.backend_data.get("pipeline_index", 0)
+
+        for idx, func in enumerate(pipeline[pipeline_index:]):
+            self.backend_data["pipeline_index"] = pipeline_index + idx
+            result = func(**self.backend_data) or {}
+            self.backend_data.update(result)
+
+    def get_pipeline(self):
+        """
+        Returns the RTI pipeline, which is a list of functions to be executed.
+        """
+        return [
+            # Add RTI pipeline functions here
+        ]
--- a/eox_nelp/pearson_vue/tasks.py
+++ b/eox_nelp/pearson_vue/tasks.py
@@ -1,0 +1,30 @@
+"""
+This module contains functions and classes for making asynchronous calls to Pearson VUE's RTI services.
+
+Functions:
+    real_time_import_task(data: dict) -> None: Performs an asynchronous call to the RTI service.
+"""
+
+from celery import shared_task
+
+from eox_nelp.pearson_vue.rti_backend import RealTimeImport
+
+
+@shared_task(bind=True)
+def real_time_import_task(self, pipeline_index=0, **kwargs):
+    """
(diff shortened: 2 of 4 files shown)


eox_nelp/pearson_vue/tasks.py
```diff
@@ -0,0 +1,30 @@
...
+        self: The Celery task instance.
+        pipeline_index (int): The index of the pipeline to be executed (default is 0).
+        **kwargs: Additional keyword arguments to configure the RTI service.
+    """
+    rti = RealTimeImport(pipeline_index=pipeline_index, **kwargs.copy())
+
+    try:
+        rti.run_pipeline()
+    except Exception as exc:  # pylint: disable=broad-exception-caught
+        self.retry(exc=exc, kwargs=rti.backend_data)
```
johanseto, line 30: max_retries or there would be a retry policy?
https://docs.celeryq.dev/en/latest/userguide/tasks.html#automatic-retry-for-known-exceptions

eox_nelp/pearson_vue/rti_backend.py
```diff
@@ -0,0 +1,49 @@
...
+            **kwargs: Additional keyword arguments to configure the RealTimeImport instance.
+        """
+        self.backend_data = kwargs.copy()
+
+    def run_pipeline(self):
+        """
+        Executes the RTI pipeline by iterating through the pipeline functions.
+        """
+        pipeline = self.get_pipeline()
+        pipeline_index = self.backend_data.get("pipeline_index", 0)
```
johanseto, line 36: why pipeline_index was selected to be a member of backend_data kwargs, instead of being a class attribute keep it as int ?
`self.pipeline_index`

johanseto: This could isolate the pipeline_index if some method change it forcing in the kwargs output of a ppeline_step.
eg:
 ``` python
def crazy_pipeline_step:
    return {"pipeline_index": 1000}
```

File: eox_nelp/pearson_vue/rti_backend.py [hunk 3]
```diff
@@ -0,0 +1,49 @@
...
+
+        for idx, func in enumerate(pipeline[pipeline_index:]):
+            self.backend_data["pipeline_index"] = pipeline_index + idx
+            result = func(**self.backend_data) or {}
```
johanseto, line 40:
```suggestion
            result = func(**self.backend_data) or {}
            self.backend_data["pipeline_index"] = pipeline_index + idx
```
may be the pipeline_index would be updated after the function works, if failsI  think that the pipeline would retry the function that fails. and  Not the next one  🤔

eox_nelp/pearson_vue/tests/test_tasks.py
```diff
@@ -0,0 +1,71 @@
...
+            "user_id": 5,
+            "course_id": "course-v1:FutureX+guide+2023"
+        }
+        second_attempt_data = {
+            "user_id": 5,
+            "course_id": "course-v1:FutureX+guide+2023",
+            "email": "admin@example.com",
+        }
+        mock_exception = Exception("Test exception")
+        self.mock_rti.run_pipeline.side_effect = [mock_exception, None]
```
johanseto, line 61: AS mentioned above, may be here is not tested what is happening with the `pipeline_index` key in the run_pipeline method.

eox_nelp/pearson_vue/tasks.py
andrey-canon: When the value is None that uses the default value https://docs.celeryq.dev/en/stable/reference/celery.app.task.html#celery.app.task.Task.retry

eox_nelp/pearson_vue/rti_backend.py
andrey-canon: Yes, that is the idea if something fails that should resume from the failure pipeline not the next one

eox_nelp/pearson_vue/tests/test_tasks.py
andrey-canon: Here is not the place since this is the test_task.py file but it's a good a idea I will add a test that checks the pipeline_index behavior

eox_nelp/pearson_vue/rti_backend.py
andrey-canon: When the task fails the retry method requires all this information, so it's easier to pass just one argument

eox_nelp/pearson_vue/rti_backend.py
johanseto: I validate with the last test and kwnowin that idx starts in zero, the position is ok to recall that position.

johanseto (approved):

PR #155, opened 2024-05-20
feat: add get_user_data pipeline 

andrey-canon:
## Description

This adds a new pipeline step that returns the basic user data

## Testing instructions

 1. Create a user and complete profile multiple fields 
 2. Go to the pytho shell `./manage.py lms shell`
 3. import the function `from eox_nelp.pearson_vue.pipeline import get_user_data`
 4. Call the function with the id of the user that was created on the step 1. `get_user_data(id)`
 5. Check that the output return the right data
image


eox_nelp/pearson_vue/tests/test_pipeline.py
```diff
@@ -0,0 +1,78 @@
...
+        """
+        Test get_user_data , this checks that the response is the expected when the
+        phone_number has and doesn't have the country code.
+
+            Expected behavior:
+            - The result is as the expected output.
+        """
+        phone_number = phone[3:] if phone.startswith("+") else phone
+        country_code = "44"  # This is the code for GB
+        self.profile.phone_number = phone
```
johanseto, line 59: If here is the phone what is the reason of a user with phone_number profile?
https://github.com/eduNEXT/eox-nelp/blob/84ae791e8e73ec21bcd4711bbba830c29921161d/eox_nelp/pearson_vue/tests/test_pipeline.py#L35

eox_nelp/pearson_vue/tests/test_pipeline.py
```diff
@@ -0,0 +1,78 @@
...
+
+        expected_output = {
+            "anonymous_user_id": anonymous_id_for_user(self.user, None),
```
johanseto, line 62: Incredible that the mock returns the same value of the expected_ouput and the called method in `get_user_data`. I didn't appreciate that

johanseto (approved):
LGTM, I suggest little change of a test if you want. Otherwise go.

eox_nelp/pearson_vue/pipeline.py
```diff
@@ -0,0 +1,66 @@
...
+            - mobile_country_code (int): The country code of the user's mobile number (same as phone_country_code).
+    """
+    user = User.objects.get(id=user_id)
+    profile = user.profile
+    phone = profile.phone_number
+    pn = phonenumbers.parse(phone) if phone.startswith("+") else phonenumbers.parse(phone, profile.country.code)
+    phone = str(pn.national_number)
+    phone_country_code = str(pn.country_code)
+
+    return {
```
johanseto, line 53: What do you think of return all this  in one dict?

andrey-canon: sorry I don't understand, this is returning a dict so what do you refer with `one dict` ?

johanseto: yes I could explain me better.
That the profile backend would receive everything but the kwargs in the pipeline would be so separated. 
I think you could return like 
``` python
return {
    "profile_metadata": {
         "username": "testicouser",
          "country": "ARG",
        ...
  }
}
```
This with the purpose to help when I create a pipe, I don't have the kwargs too split and then,  kwargs.get(country)...
I would have all the profile metadata in one dict.
I only want your opinion...

eox_nelp/pearson_vue/pipeline.py
```diff
@@ -51,16 +51,18 @@ def get_user_data(user_id, **kwargs):  # pylint: disable=unused-argument
...
-        "last_name": user.last_name,
-        "email": user.email,
-        "address": profile.mailing_address,
-        "city": profile.city,
-        "country": profile.country.code,
-        "phone_number": phone,
-        "phone_country_code": phone_country_code,
-        "mobile_number": phone,
-        "mobile_country_code": phone_country_code,
+        "profile_metadata": {
```
johanseto, line 54: [👍]

johanseto (approved):

johanseto (commented):
LGTM, missing only a rebase =)

PR #158, opened 2024-05-21
Jlc/pearson vue/connect completion grading receivers

johanseto:
## Description

feat: configure receiver and pipe for completion and grade passed course.

Created 2 receivers one for block_completion post_save signal, and the other  `COURSE_GRADE_NOW_PASSED` signal.
Each one has a setting to manage.
- `PEARSON_RTI_ACTIVATE_COMPLETION_GATE`
- `PEARSON_RTI_ACTIVATE_GRADED_GATE`

Then is a pipeline(`handle_course_completion_status`) that analyzes whether to continue the pipeline in completion cases or not. This would be the first on the list.
The pipeline has 3 behaviors depending on the case:
  - is_passing is true means the course is graded(passed) and doesn't need this pipe validation.
    The pipeline continues without changes.
  - is_complete=True and is_graded=False pipeline should continue with uploaded metadata
          (completed courses and not graded).
  - Otherwise, the pipeline will stop, for graded-courses the `COURSE_GRADE_NOW_PASSED` signal would act with (`is_passing`).
## Testing instructions

Check test of the pipeline.

## Checklist for Merge

- [ ] Updated documentation
- [ ] Rebased master/main
- [ ] Squashed commits

### DIFF
--- a/eox_nelp/pearson_vue/pipeline.py
+++ b/eox_nelp/pearson_vue/pipeline.py
@@ -25,7 +25,7 @@ logger = logging.getLogger(__name__)
 User = get_user_model()
 
 
-def terminate_not_full_completion_cases(user_id, course_id, **kwargs):
+def handle_course_completion_status(user_id, course_id, **kwargs):
     """Pipeline that check the case of completion cases on the pipeline execution. Also this pipe
     has 4 behaviours depending the case:
         - skip this pipeline if setting PEARSON_RTI_TESTING_SKIP_CHECK_COMPLETION is truthy. Pipeline continues.
@@ -33,7 +33,8 @@ def terminate_not_full_completion_cases(user_id, course_id, **kwargs):
           The pipeline continues without changes.
         - is_complete=True and is_graded=False pipeline should continue.
           (completed courses and not graded).
-        - Otherwise the pipeline will stop, for grading-courses the COURSE_GRADE_NOW_PASSED signal would act.
+        - Otherwise this indicates that the pipeline execution would be stopped,
+          for grading-courses the COURSE_GRADE_NOW_PASSED signal would act.
 
     Args:
         user_id (int): The ID of the user whose data is to be retrieved.
@@ -43,9 +44,9 @@ def terminate_not_full_completion_cases(user_id, course_id, **kwargs):
     Returns:
         dict: Pipeline dict
     """
-    if getattr(settings, "PEARSON_RTI_TESTING_SKIP_FULL_COMPLETION_CASES", False):
+    if getattr(settings, "PEARSON_RTI_TESTING_SKIP_HANDLE_COURSE_COMPLETION_STATUS", False):
         logger.info(
-            "Skipping `terminate_not_full_completion_cases` pipe for user_id:%s and course_id: %s",
+            "Skipping `handle_course_completion_status` pipe for user_id:%s and course_id: %s",
             str(user_id),
             course_id
         )
--- a/eox_nelp/pearson_vue/pipeline.py
+++ b/eox_nelp/pearson_vue/pipeline.py
@@ -25,7 +25,7 @@ logger = logging.getLogger(__name__)
 User = get_user_model()
 
 
-def check_completion_metadata(user_id, course_id, **kwargs):
+def terminate_not_full_completion_cases(user_id, course_id, **kwargs):
     """Pipeline that check the case of completion cases on the pipeline execution. Also this pipe
     has 3 behaviours depending the case:
         - is_passing is true means the course is graded(passed) and dont needs this pipe validation.
@@ -48,10 +48,7 @@ def check_completion_metadata(user_id, course_id, **kwargs):
     is_complete, is_graded = get_completed_and_graded(user_id, course_id)
 
     if is_complete and not is_graded:
-        return {
-            "is_complete": is_complete,
-            "is_graded": is_graded,
-        }
+        return None
 
     return {
         "safely_pipeline_termination": True,
--- a/eox_nelp/pearson_vue/pipeline.py
+++ b/eox_nelp/pearson_vue/pipeline.py
@@ -27,10 +27,11 @@ User = get_user_model()
 
 def terminate_not_full_completion_cases(user_id, course_id, **kwargs):
     """Pipeline that check the case of completion cases on the pipeline execution. Also this pipe
-    has 3 behaviours depending the case:
+    has 4 behaviours depending the case:
+        - skip this pipeline if setting PEARSON_RTI_TESTING_SKIP_CHECK_COMPLETION is truthy. Pipeline continues.
         - is_passing is true means the course is graded(passed) and dont needs this pipe validation.
           The pipeline continues without changes.
-        - is_complete=True and is_graded=False pipeline should continue with uploaded metadata
+        - is_complete=True and is_graded=False pipeline should continue.
           (completed courses and not graded).
         - Otherwise the pipeline will stop, for grading-courses the COURSE_GRADE_NOW_PASSED signal would act.
 
@@ -42,6 +43,14 @@ def terminate_not_full_completion_cases(user_id, course_id, **kwargs):
     Returns:
         dict: Pipeline dict
     """
+    if getattr(settings, "PEARSON_RTI_TESTING_SKIP_FULL_COMPLETION_CASES", False):
+        logger.info(
+            "Skipping `terminate_not_full_completion_cases` pipe for user_id:%s and course_id: %s",
+            str(user_id),
+            course_id
+        )
+        return None
+
     if kwargs.get("is_passing"):
         return None
 
--- a/eox_nelp/pearson_vue/pipeline.py
+++ b/eox_nelp/pearson_vue/pipeline.py
@@ -28,7 +28,8 @@ User = get_user_model()
 def handle_course_completion_status(user_id, course_id, **kwargs):
     """Pipeline that check the case of completion cases on the pipeline execution. Also this pipe
     has 4 behaviours depending the case:
-        - skip this pipeline if setting PEARSON_RTI_TESTING_SKIP_CHECK_COMPLETION is truthy. Pipeline continues.
+        - skip this pipeline if setting PEARSON_RTI_TESTING_SKIP_HANDLE_COURSE_COMPLETION_STATUS is truthy.
+          Pipeline continues.
         - is_passing is true means the course is graded(passed) and dont needs this pipe validation.
           The pipeline continues without changes.
         - is_complete=True and is_graded=False pipeline should continue.
--- a/eox_nelp/pearson_vue/pipeline.py
+++ b/eox_nelp/pearson_vue/pipeline.py
@@ -7,6 +7,8 @@ called sequentially, where each function processes data and passes it along to t
 Functions:
     get_user_data(data: dict) -> dict: Retrieves and processes user data.
 """
+import logging
+
 import phonenumbers
 from django.conf import settings
 from django.contrib.auth import get_user_model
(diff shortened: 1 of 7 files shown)


eox_nelp/pearson_vue/pipeline.py
```diff
@@ -0,0 +1,104 @@
...
+
+    if is_complete and not is_graded:
+        return {
+            "is_complete": is_complete,
+            "is_graded": is_graded,
+        }
+
+    return {
+        "safely_pipeline_termination": True,
+    }
```
andrey-canon, line 104:
```suggestion
    is_complete, is_graded = get_completed_and_graded(user_id, course_id)
    
    if kwargs.get("is_passing") or (is_complete and not is_graded):
        return {}

    return {
        "safely_pipeline_termination": True,
    }
```

andrey-canon (changes requested):

eox_nelp/pearson_vue/pipeline.py
johanseto: @andrey-canon I was thinking in this and I could skip the `get_completed_and_graded` call for grading course. That method uses modulestore,  so it skips a call in performance. 
Thinking more the better performance  approach is using  only the receiver of `COURSE_GRADE_NOW_PASSED` in order, to skip a lot of modulestore call thinking the completion of each unit, of each user, of each course =O xD. 
But we could still have both approaches

eox_nelp/pearson_vue/pipeline.py
```diff
@@ -16,10 +18,46 @@
...
+
+    Args:
+        user_id (int): The ID of the user whose data is to be retrieved.
+        course_id (str): course_id to check completion or graded.
+        **kwargs: Additional keyword arguments.
+
+    Returns:
+        dict: Pipeline dict
+    """
+    if kwargs.get("is_passing"):
```
johanseto, line 45: @andrey-canon I was thinking in this and I could skip the get_completed_and_graded call for grading course without the suggestion. That method uses modulestore, so it skips a call in performance.
Thinking more the better performance approach is using only the receiver of COURSE_GRADE_NOW_PASSED in order, to skip a lot of modulestore call thinking the completion of each unit, of each user, of each course =O xD.
But we could still have both approaches

eox_nelp/pearson_vue/pipeline.py
andrey-canon: But now this does't works  the None that you return is no compatible with the update method, so this will raise an exception when is_passing is True,  indeed now i don't like the suggestion since I looking for 
congruence in the  output  and you made me think that if we don't need anything when  is_passing is true  we don't need  is_complete and  is_graded at all, now this is my suggestion

```python
    safely_pipeline_termination = True
        
    if kwargs.get("is_passing"):
        safely_pipeline_termination = False 
    else:
        is_complete, is_graded = get_completed_and_graded(user_id, course_id)
        safely_pipeline_termination = not(is_complete and not is_graded)

    return {
        "safely_pipeline_termination": safely_pipeline_termination,
    }
```

johanseto: I think it could be compatible with the None response? the empty dict is overriding.
https://github.com/eduNEXT/eox-nelp/blob/e9f93d1bebddcdfe5e79530d289c16fa6c5fad7b/eox_nelp/pearson_vue/rti_backend.py#L41

WHat about:
```python3
    if kwargs.get("is_passing"):
        return 
    
    is_complete, is_graded = get_completed_and_graded(user_id, course_id)
    if is_complete and not is_graded:
        return

    return {
        "safely_pipeline_termination": True,
    }
```

eox_nelp/pearson_vue/rti_backend.py
```diff
@@ -50,5 +50,6 @@ def get_pipeline(self):
         Returns the RTI pipeline, which is a list of functions to be executed.
         """
         return [
+            check_completion_metadata,
```
andrey-canon, line 53: could you rebase and include the others pipes get_user_data get_exam_data, .....

eox_nelp/pearson_vue/pipeline.py
andrey-canon: pylint error if I'm not wrong

johanseto: So with return None, you dont like it or empty dict. Some think is that the nextes pipe would have the safely_pipeline_termination kwarg in False defined

johanseto: Check [feacd71](https://github.com/eduNEXT/eox-nelp/pull/158/commits/feacd712aadc2e7cffdde9cbb938c76840b824e3)

eox_nelp/pearson_vue/pipeline.py
```diff
@@ -16,10 +18,43 @@
...
 
 User = get_user_model()
 
 
+def terminate_not_full_completion_cases(user_id, course_id, **kwargs):
+    """Pipeline that check the case of completion cases on the pipeline execution. Also this pipe
+    has 3 behaviours depending the case:
+        - is_passing is true means the course is graded(passed) and dont needs this pipe validation.
+          The pipeline continues without changes.
+        - is_complete=True and is_graded=False pipeline should continue with uploaded metadata
```
andrey-canon: I think that was change, I mean there is no metadata

eox_nelp/pearson_vue/pipeline.py
```diff
@@ -16,10 +18,43 @@
...
 from eox_nelp.pearson_vue.constants import PAYLOAD_CDD, PAYLOAD_EAD, PAYLOAD_PING_DATABASE
 from eox_nelp.pearson_vue.utils import update_xml_with_dict
+from eox_nelp.signals.utils import get_completed_and_graded
+
+logger = logging.getLogger(__name__)
 
 User = get_user_model()
 
 
+def terminate_not_full_completion_cases(user_id, course_id, **kwargs):
```
andrey-canon, line 28: ChatGPT suggest that this name `handle_course_completion_status` describes better the functionality

eox_nelp/pearson_vue/pipeline.py
```diff
@@ -16,10 +18,43 @@
...
+    """Pipeline that check the case of completion cases on the pipeline execution. Also this pipe
+    has 3 behaviours depending the case:
+        - is_passing is true means the course is graded(passed) and dont needs this pipe validation.
+          The pipeline continues without changes.
+        - is_complete=True and is_graded=False pipeline should continue with uploaded metadata
+          (completed courses and not graded).
+        - Otherwise the pipeline will stop, for grading-courses the COURSE_GRADE_NOW_PASSED signal would act.
```
andrey-canon, line 36:
```suggestion
        - Otherwise this indicates that the pipeline execution could be stopped, for grading-courses the COURSE_GRADE_NOW_PASSED signal would act.
```

??

eox_nelp/pearson_vue/tests/test_pipeline.py
```diff
@@ -19,11 +19,134 @@
...
+
+        get_completed_and_graded_mock.assert_not_called()
+        self.assertIsNone(result)
+
+    @patch("eox_nelp.pearson_vue.pipeline.get_completed_and_graded")
+    def test_check_is_complete_not_graded(self, get_completed_and_graded_mock):
+        """Test the pipeline return expected dict with empty `is_passing`, and
+           is_complete=True, is_graded=False.
+        Expected behavior:
+            - Returned value is dict with is_complete and is_graded.
```
andrey-canon, line 86: Please update it

eox_nelp/pearson_vue/tests/test_pipeline.py
```diff
@@ -19,11 +19,134 @@
...
+    @patch("eox_nelp.pearson_vue.pipeline.get_completed_and_graded")
+    def test_check_complete_graded(self, get_completed_and_graded_mock):
+        """Test the pipeline return expected dict(safely_pipeline_termination) with empty `is_passing`, and
+           is_complete=True, is_graded=True.
+        Expected behavior:
+            - Returned value is dict with `safely_pipeline_termination`
+        """
+        pipeline_kwargs = {}
+        get_complete_and_graded_output = {
+            "is_complete": False,
```
andrey-canon, line 160:
```suggestion
            "is_complete": True,
```

Based on the docstring

eox_nelp/pearson_vue/tests/test_pipeline.py
```diff
@@ -19,11 +19,134 @@
...
+            "safely_pipeline_termination": True,
+        }
+        get_completed_and_graded_mock.return_value = (
+            get_complete_and_graded_output["is_complete"],
+            get_complete_and_graded_output["is_graded"],
+        )
+
+        result = terminate_not_full_completion_cases(self.user_id, self.course_id, **pipeline_kwargs)
+
+        self.assertDictEqual(expected_output, result)
```
andrey-canon, line 173:
```suggestion
        self.assertIsNone(result)
```
That happens with  is_complete=True and is_graded=True

eox_nelp/pearson_vue/tests/test_pipeline.py
```diff
@@ -19,11 +19,134 @@
...
+
+    @patch("eox_nelp.pearson_vue.pipeline.get_completed_and_graded")
+    def test_check_complete_graded(self, get_completed_and_graded_mock):
```
andrey-canon, line 152: This test makes the same that the previous one

andrey-canon (changes requested):

eox_nelp/pearson_vue/tests/test_pipeline.py
```diff
@@ -37,12 +38,37 @@ def setUp(self):
...
+    def test_skip_pipe_with_settings(self, get_completed_and_graded_mock):
+        """Test the pipeline is skipped with truthy
+        `PEARSON_RTI_TESTING_SKIP_FULL_COMPLETION_CASES` setting.
+        Expected behavior:
+            - logger info message expected
+            - get_completed_and_graded_mock is not called.
+            - Returned value is None
+
+        """
+        setattr(settings, "PEARSON_RTI_TESTING_SKIP_FULL_COMPLETION_CASES", True)
```
andrey-canon, line 52: Since this is a simple setting you can include that in the decorator

eox_nelp/pearson_vue/pipeline.py
andrey-canon: You have already changed that but I'm not sure the next line, that makes sense ?

eox_nelp/pearson_vue/tests/test_pipeline.py
johanseto: I think here is dict with safely termination

eox_nelp/pearson_vue/pipeline.py
johanseto: 302e8de11d0229eed849a0f45998648ce089ef3b

eox_nelp/pearson_vue/tests/test_pipeline.py
andrey-canon: you'r right because it's `not is_graded`

andrey-canon (approved):

eox_nelp/pearson_vue/pipeline.py
```diff
@@ -16,10 +18,53 @@
...
+
+logger = logging.getLogger(__name__)
 
 User = get_user_model()
 
 
+def handle_course_completion_status(user_id, course_id, **kwargs):
+    """Pipeline that check the case of completion cases on the pipeline execution. Also this pipe
+    has 4 behaviours depending the case:
+        - skip this pipeline if setting PEARSON_RTI_TESTING_SKIP_CHECK_COMPLETION is truthy. Pipeline continues.
```
andrey-canon, line 31: You changed the name of the setting however you didn't update the docstring

PR #160, opened 2024-05-21
fix: include inner wrapper to avoid  JSON serializable errors

andrey-canon:
## Description

This includes a inner method into auditable methods to avoid  JSON serializable errors

## Additional information

## Checklist for Merge

- [ ] Updated documentation
- [ ] Rebased master/main
- [ ] Squashed commits

### DIFF
--- a/eox_nelp/api_clients/pearson_rti.py
+++ b/eox_nelp/api_clients/pearson_rti.py
@@ -62,7 +62,6 @@ class PearsonRTIApiClient(AbstractSOAPClient):
 
         return result
 
-    @audit_method(action="Import Candidate Demographics")
     def import_candidate_demographics(self, payload):
         """The CDD service imports candidate demographic data in to the VUE system. This
         method performs the required action to store the data into the Pearson service.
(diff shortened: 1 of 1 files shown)


eox_nelp/api_clients/pearson_rti.py
```diff
@@ -62,7 +62,6 @@ def ping(self, payload):
 
         return result
 
-    @audit_method(action="Import Candidate Demographics")
```
johanseto, line 65: So basicly, the problem is that could be serialized the error because is a function attr of the  PearsonRTIApiClient class.?

johanseto (approved):
I think we can move on

eox_nelp/api_clients/pearson_rti.py
andrey-canon: eox-audit takes as parameter the self declaration, in this case self is an instance of PearsonRTIApiClient and that class is not serializable, the inner method removes the self from the eox-audit arguments

PR #164, opened 2024-05-22
feat: adds a a new model and multiple views that allow to save the re…

andrey-canon:
## Description

This adds a basic model with multiple view that exposes the pearson rten service


eox_nelp/migrations/0007_auto_20240522_2251.py
```diff
@@ -0,0 +1,22 @@
+# Generated by Django 3.2.21 on 2024-05-22 22:51
```
johanseto, line 1: Could you change the name of the migration to something more telling(verbose)

eox_nelp/pearson_vue/api/v1/urls.py
```diff
@@ -0,0 +1,36 @@
...
+    ResultNotificationView,
+    RevokeResultView,
+    UnrevokeResultView,
+)
+
+app_name = "eox_nelp"  # pylint: disable=invalid-name
+
+urlpatterns = [
+    path('resultNotification/', ResultNotificationView.as_view(), name='resultNotification'),
+    path('placeHold/', PlaceHoldView.as_view(), name='placeHold'),
```
johanseto, line 31: Why the name is in camelcase? url is ok to be in camel but the nameurl to use reverse?

eox_nelp/pearson_vue/api/v1/views.py
```diff
@@ -0,0 +1,139 @@
...
+from edx_rest_framework_extensions.auth.jwt.authentication import JwtAuthentication
+from rest_framework import generics, status
+from rest_framework.permissions import IsAuthenticated
+from rest_framework.response import Response
+
+from eox_nelp.pearson_vue.api.v1.serializers import PearsonRTENSerializer
+from eox_nelp.pearson_vue.models import PearsonRTENModel
+
+
+class PearsonRTENBaseView(generics.CreateAPIView):
```
johanseto, line 24: What do you think of the [ListCreateAPIView](https://www.django-rest-framework.org/api-guide/generic-views/#listcreateapiview)
I mean in order to view or look via API?

eox_nelp/pearson_vue/api/v1/views.py
```diff
@@ -0,0 +1,139 @@
...
+
+class PlaceHoldView(PearsonRTENBaseView):
+    """
+    View for handling Place Hold events.
+
+    This view handles the creation of Place Hold events in the Pearson RTEN system.
+    The `event_type` attribute is set to "placeHold".
+    """
+
+    event_type = "placeHold"
```
johanseto, line 95: Could we create something like constant and load for that something like 
https://github.com/eduNEXT/eox-nelp/pull/164/files#diff-c9d49fedc8ccd5477e993f68a6696079ac46a3fbceb64f6f7e3af9c8f5d4ccf1R15
I mean also the test use the event type, and seem has the same shape in URLs to give the name.

eox_nelp/migrations/0007_auto_20240522_2251.py
andrey-canon: Done, just keep in mind that those files are autogenerated, the name is the django standard, in this case  this depends on [this](https://github.com/eduNEXT/eox-nelp/pull/166)  so the new migration has a better name

eox_nelp/pearson_vue/api/v1/urls.py
andrey-canon: Changed in the constant implementation https://github.com/eduNEXT/eox-nelp/pull/164/commits/98fb31ce2d16c6af30c5a8708e579d432f4d272b

eox_nelp/pearson_vue/api/v1/views.py
andrey-canon: Changed in the constants implementation https://github.com/eduNEXT/eox-nelp/pull/164/commits/98fb31ce2d16c6af30c5a8708e579d432f4d272b

eox_nelp/pearson_vue/api/v1/views.py
andrey-canon: Added https://github.com/eduNEXT/eox-nelp/pull/164/commits/59da7171efd1ee78149b5b22789b1e53166ee79d

eox_nelp/pearson_vue/api/v1/views.py
```diff
@@ -0,0 +1,158 @@
...
+        permission_classes (list): List of permission classes required for the view.
+        serializer_class (Serializer): Serializer class used for validating and parsing data.
+        queryset (QuerySet): Queryset representing the model instances to operate on.
+        authentication_classes (list): List of authentication classes used for authentication.
+    """
+    event_type = None
+    permission_classes = [IsAuthenticated]
+    serializer_class = PearsonRTENSerializer
+    queryset = PearsonRTENModel.objects.all()  # pylint: disable=no-member
+    authentication_classes = [JwtAuthentication]
```
johanseto, line 49: What do you think of adding session authentication. And staff or admin permission?. This to allow check the API via webrowser?
image

johanseto (commented):
In form is possible to change the word model to event?
image

If is too much work avoid this item and continue with model.

eox_nelp/pearson_vue/api/v1/views.py
andrey-canon: That requires  that the credentials that we will share with Pearson are associated with an staff or admin user, and we won't give admin credentials to them.

on the other hand, the session authentication doesn't make sense for me, since this depends on a certificate, probably I will  do this more restrictive and include the Bearer authentication that I found in eox-tenant

johanseto (approved):
LGTM. Local env Bearer tested and post ResultNotification -> 200 with empty response.
Peek 2024-06-04 10-41

PR #167, opened 2024-05-31
feat: this includes the client authorization id into the EAD request

andrey-canon:
## Description

This add the client authorization id the EAD request
image

## Additional information

## Checklist for Merge

- [ ] Updated documentation
- [ ] Rebased master/main
- [ ] Squashed commits

### DIFF
--- a/eox_nelp/pearson_vue/pipeline.py
+++ b/eox_nelp/pearson_vue/pipeline.py
@@ -5,7 +5,12 @@ Each function in this module is designed to perform a specific step in the pipel
 called sequentially, where each function processes data and passes it along to the next step in the pipeline.
 
 Functions:
-    get_user_data(data: dict) -> dict: Retrieves and processes user data.
+    handle_course_completion_status: Pipeline function to handle course completion status.
+    get_user_data: Retrieves and processes user data.
+    check_service_availability: Checks the availability of the Pearson VUE RTI service.
+    import_candidate_demographics: Imports candidate demographics data.
+    import_exam_authorization: Imports exam authorization data.
+    get_exam_data: Retrieves exam data.
 """
 import logging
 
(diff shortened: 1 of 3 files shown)


eox_nelp/pearson_vue/pipeline.py
```diff
@@ -292,7 +293,7 @@ def import_exam_authorization(
         raise Exception("Error trying to process import exam authorization request.")
 
 
-def get_exam_data(course_id, **kwargs):  # pylint: disable=unused-argument
+def get_exam_data(user_id, course_id, **kwargs):  # pylint: disable=unused-argument
     """
     Retrieves exam data from Django settings.
```
johanseto, line 299: Could you update the docstring with user_id?

johanseto:
@andrey-canon I was testing and noticed that for anonymous_ids:
 - `None` key is related to the `CDD(clientCandidateId)` 
- `{course_id}` key related   to the `EAD(clientAuthorizationId)`
image

eox_nelp/pearson_vue/pipeline.py
```diff
@@ -314,6 +315,13 @@ def get_exam_data(course_id, **kwargs):  # pylint: disable=unused-argument
     """
     courses_data = getattr(settings, "PEARSON_RTI_COURSES_DATA")
     exam_metadata = courses_data[course_id]
+
+    # This generates the clientAuthorizationID based on the user_id and course_id
+    exam_metadata["client_authorization_id"] = anonymous_id_for_user(
```
johanseto, line 320:
```suggestion
    exam_metadata["@client_authorization_id"] = anonymous_id_for_user(
```
What do you think of adding the `@` char to know that is related and arg of tag? and not tag? of ead

andrey-canon: current implementation uses the `get_exam_data` result inside `import_exam_authorization ` pipe, probably this makes sense in that context, however if we consider the `get_exam_data`  method out of the context the `@` doesn't make sense and the result of this could be used in other pipes or methods which don't use the soap notation

johanseto (approved):

PR #168, opened 2024-06-04
And/refactor phone validation

andrey-canon:
## Description
This makes a  refactor of the phone_validation feature,  basically this splits the current structure into two folders, on-time-password(Contains all the related code with the otp validation and generation) and user-profile  that contains all the code that allows to update the user-profile records

### Before
/eox-nelp/api/phone-validation/v1/generate-otp/
/eox-nelp/api/phone-validation/v1/validate-otp/

### After

/eox-nelp/api/one-time-password/v1/generate-otp/
/eox-nelp/api/user-profile/v1/update-user-data/

## Additional information

## Checklist for Merge

- [ ] Updated documentation
- [ ] Rebased master/main
- [ ] Squashed commits


johanseto (commented):
I am thinking of the implementation and the idea is to force the validate otp with the update profile metadata. But there is no way to update profile metadata without the otp validation, because I think phone validation would manage only the phone field. Why the other fields needs the phone to be updated?

andrey-canon:
> Why the other fields needs the phone to be updated?

Because we need the phone number to send the OTP
Anyway I don't understand how this concern is related with this PR, this is basically a refactor,  I know the implementation has changed, urls were modified but the functionality has not changed at all, so at this stage I'm expecting a review that is related with the PR scope and your comment , it's totally valid, but I feel it's out of the scope

eox_nelp/user_profile/api/v1/views.py
```diff
@@ -0,0 +1,79 @@
...
+@api_view(["POST"])
+@authentication_classes((
+    SessionAuthenticationAllowInactiveUser,
+))
+@permission_classes((IsAuthenticated,))
+def update_user_data(request):
+    """ View to update user's fields.
+    ## Usage
+
+    ### **POST** /eox-nelp/api/user-profile/v1/update-user-data/
```
johanseto, line 34: This is an update view and works with a Post request? Does it have to be that way?

eox_nelp/user_profile/api/v1/views.py
```diff
@@ -0,0 +1,79 @@
...
+        )
+
+    user_otp_key = f"{request.user.username}-{user_phone_number}"
+    logger.info("validating otp for %s*****", user_otp_key[:-5])
+
+    if not proposed_user_otp == cache.get(user_otp_key):
+        return HttpResponseForbidden(reason="Forbidden - wrong code")
+
+    try:
+        with transaction.atomic():
```
johanseto, line 66: Why do you need to be atomic?

eox_nelp/user_profile/api/v1/urls.py
```diff
@@ -0,0 +1,10 @@
+"""eox_nelp puser_profile.api v1 urls"""
+from django.urls import path
+
+from eox_nelp.user_profile.api.v1.views import update_user_data
+
+app_name = "eox_nelp"  # pylint: disable=invalid-name
+
+urlpatterns = [
+    path("update-user-data/", update_user_data, name="update-user-data"),
```
johanseto, line 9: it could be
```suggestion
    path("user-data/", update_user_data, name="update-user-data"),
```
Then with HTTP verbs control the action to create, update, delete or is a very different think?

eox_nelp/user_profile/api/v1/views.py
johanseto: Why do you select the atomic logic in context manager and not in decorator way to use?
https://docs.djangoproject.com/en/3.2/topics/db/transactions/#controlling-transactions-explicitly

eox_nelp/user_profile/api/v1/views.py
andrey-canon: I think you implemented that, as mentioned before this is a refactor and I keep same functionality, you can check the first commit where github just marked the renamed file https://github.com/eduNEXT/eox-nelp/pull/168/commits/3e2a95981c552ab39405c25fe4be48bdbc86733f#diff-a79e9a949c512eb3de51c9ca381be97e75441032280b46f251fdd4118f8d6f1a

eox_nelp/user_profile/api/v1/views.py
andrey-canon: You can find the answer of the first question in this [link ](https://github.com/nelc/edx-platform/blob/open-release/palm.nelp/openedx/core/djangoapps/user_api/accounts/api.py#L130)

same thing different implementation, if you prefer one of them please expose your arguments

eox_nelp/user_profile/api/v1/urls.py
andrey-canon: Since the unique allowed verb is post the name helps to identify the view purpose

eox_nelp/user_profile/api/v1/views.py
johanseto: I am ok, I only want to know if there was a particular reason. Anyway, no problem.

eox_nelp/user_profile/api/v1/views.py
johanseto: Yes but I didnt use an endpoint with `update` wording. But for the moment,I will not delay with the forms

johanseto (approved):

johanseto (commented):
Finally, as a suggestion what do you think of rename the module
image
to `otp`?

eox_nelp/user_profile/api/v1/views.py
andrey-canon: But I had to use the update word because this uses a POST instead a PUT hahaha to be honest I tried to implement the PUT however the OTP mixin doesn't work with PUT, so changing that requires more things that a simple refactor

andrey-canon:
> Finally, as a suggestion what do you think of rename the module image to `otp`?

I prefer to be explicit since OTP could have another meaning

eox_nelp/user_profile/api/v1/views.py
johanseto: Now it is very very clear. 
  I prefer how you treated it and avoid  giving more complexity to the code with a PUT

PR #169, opened 2024-06-04
feat: user course enrollment for clientAuthID

johanseto:
# Description
This is for pearson RTI ead request

## Description

Describe what this pull request changes, and why. Include implications for people using this change.

Useful information to include:
- Include screenshots for changes to the UI (ideally, both "before" and "after" screenshots, if applicable).
- Provide links to the description of corresponding configuration changes. Remember to correctly annotate these
changes.

## Additional information

## Checklist for Merge

- [ ] Updated documentation
- [ ] Rebased master/main
- [ ] Squashed commits

### DIFF
--- a/eox_nelp/pearson_vue/utils.py
+++ b/eox_nelp/pearson_vue/utils.py
@@ -5,6 +5,9 @@ This includes xml helpers:
 import xmltodict
 from pydantic.v1.utils import deep_update
 
+from eox_nelp.edxapp_wrapper.student import CourseEnrollment
+from eox_nelp.pearson_vue.constants import CLIENT_AUTHORIZATION_ID_OFFSET
+
 
 def update_xml_with_dict(xml: str, update_dict: dict) -> str:
     """Update an xml string using its dict representation, deep_update,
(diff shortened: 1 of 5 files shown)


eox_nelp/pearson_vue/utils.py
```diff
@@ -26,3 +29,16 @@ def update_xml_with_dict(xml: str, update_dict: dict) -> str:
...
+
+def generate_client_authorization_id(user_id: int, course_id: str) -> str:
+    """Use course_enrollment_id to generate client_authorization_id for EAD requests.
+    Args:
+        user_id (int): user id
+        course_id (str): course id
+    Returns:
+        str: string that represents the client_auth_id
+    """
+    course_enrollment = CourseEnrollment.objects.get(user_id=user_id, course_id=course_id)
```
andrey-canon, line 42: This is a simple solution and I like that however this strongly depends on the course enrollment id and that makes me wonder what will happen if one day we need to delete  a course enrollment record to perform a support operation, if we create the enrollment again the  id will be different and we won't be able to track previous records

johanseto: You have a good point, if the course-enrollment is deleted we could not track it. 
One solution is to not make hard deletes of course enrollments and only use the `is_active` flag.
image
If not I could create an id with
```python

client_authorization_id =  f"{user_id}-{org}+{course}+{run}"
```
something like :
image

andrey-canon (approved):

PR #172, opened 2024-06-04
feat: set default phone number

andrey-canon:
## Description

This makes optional the phone_number API parameter, by taking as default option the database information

## Testing instructions
1. Set the user phone number
2. Make a POST request with an empty body `/eox-nelp/api/one-time-password/v1/generate/`
3. Check logs and response

image
image

### DIFF
--- a/eox_nelp/tests/mixins.py
+++ b/eox_nelp/tests/mixins.py
@@ -7,6 +7,7 @@ Classes:
 from ddt import data, ddt
 from django.contrib.auth import get_user_model
 from django.urls import reverse
+from mock import Mock
 from rest_framework import status
 from rest_framework.test import APIClient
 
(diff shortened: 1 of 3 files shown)


eox_nelp/tests/mixins.py
```diff
@@ -26,6 +27,7 @@ def setUp(self):  # pylint: disable=invalid-name
         self.client = APIClient()
         self.user, _ = User.objects.get_or_create(username="vader")
         self.client.force_authenticate(self.user)
+        setattr(self.user, "profile", Mock(phone_number=None))
```
johanseto, line 30: Do you think it could work with that?
```suggestion
        setattr(self.user, "profile", Mock(phone_number=""))
```

johanseto: I was confused. I think that empty in django profile field would be "". But no is None.
image

johanseto (approved):

johanseto:
@andrey-canon do you think also validate could get the phone number from request user.profile.phone???
if is not sent?
image

andrey-canon:
> @andrey-canon do you think also validate could get the phone number from request user.profile.phone??? if is not sent? image

I'm not sure if I'm understanding because this is doing that

johanseto:
> > @andrey-canon do you think also validate could get the phone number from request user.profile.phone??? if is not sent? image
> 
> I'm not sure if I'm understanding because this is doing that

Yes you are right. I was confused because in the views file is for `generate` and for `validate` is inside the decorator.

johanseto (approved):

johanseto (approved):

PR #175, opened 2024-06-13
feat: support missing user field in update_user_data API view

andrey-canon:
## Description
support missing user field in update_user_data API view

## Testing instructions
1. Add settings e,g
```python
EXTRA_ACCOUNT_USER_FIELDS = (
    "first_name",
    "last_name",
)
```
2. Go to /eox-nelp/api/user-profile/v1/update-user-data/
3. Send the field and check

ausf

## Additional information

## Checklist for Merge

- [ ] Updated documentation
- [ ] Rebased master/main
- [ ] Squashed commits

### DIFF
--- a/eox_nelp/user_profile/api/v1/views.py
+++ b/eox_nelp/user_profile/api/v1/views.py
@@ -8,6 +8,7 @@ Function-views:
 """
 import logging
 
+from django.conf import settings
 from django.db import transaction
 from edx_rest_framework_extensions.auth.session.authentication import SessionAuthenticationAllowInactiveUser
 from rest_framework import status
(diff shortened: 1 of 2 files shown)


andrey-canon (commented):
@johanseto This is the simplest way that I found to allow to edit those fields, however I'm not totally happy with that if you have more ideas please let me know

johanseto (commented):
To avoid the monkey patch...
I think in this file,
https://github.com/eduNEXT/eox-nelp/blob/cdd501c828ddc8af859489a1e9eab5b9ea3a6458/eox_nelp/user_profile/api/v1/views.py#L50-L63
we could do something like this.
``` python
    concrete_user_fields = [field.name for field in User._meta.get_fields() if field.concrete]
    for field, value in request.data.items():
        if field in concrete_user_fields:
            setattr(request.user, field, value)
    request.user.save()
```

Instead of concrete_fields we could use the `"EXTRA_ACCOUNT_USER_SERIALIZER_FIELDS"` setting.
image

Anyway  that code for user, is similar how social details pipeline updates user.

https://github.com/python-social-auth/social-core/blob/master/social_core/pipeline/user.py#L121C8-L121C35

eox_nelp/user_profile/api/v1/views.py
```diff
@@ -50,6 +51,19 @@ def update_user_data(request):
...
+
+            # This extra code block is required since the method update_account_settings just
+            # allows to update fields defined in the AccountUserSerializer and the AccountLegacyProfileSerializer
+            # so some fields like first_name and last_name are not editable in the standad implementation.
+            extra_account_user_fields = getattr(settings, "EXTRA_ACCOUNT_USER_FIELDS", [])
+
+            if extra_account_user_fields:
+                for field, value in request.data.items():
+                    if field in extra_account_user_fields:
+                        setattr(request.user, field, value)
```
johanseto, line 63:
```suggestion
                        if hasattr(request.user, field):
                            setattr(request.user, field, value)
```
What do you think of this if check, in case the setting has a wrong value like:
`EXTRA_ACCOUNT_USER_FIELDS = ["cpu_brand"]`
the object would not be modified with that attribute.

andrey-canon: The change was included in the line 62 however I made an amend instead of a commit

johanseto (approved):
LGTM

PR #181, opened 2024-06-20
feat: append id of the AnoymousUserId instance to client authorization.

andrey-canon:
## Description

The requirement is to generate an id shorter than 25 characters  that allows to associate a user with a course, the current CourseEnrollment implementation does that, however if the record is deleted the association will be lost, so this includes the id of the AnonymousUserId instance, so if the course enrollment record is deleted the id could help to identify the user-course association 

Basically this modifies the method `generate_client_authorization_id` in order to append the AnonymousUserId instance id to the current value.

## Additional information

## Checklist for Merge

- [ ] Updated documentation
- [ ] Rebased master/main
- [ ] Squashed commits


eox_nelp/pearson_vue/utils.py
```diff
@@ -40,5 +40,7 @@ def generate_client_authorization_id(user_id: int, course_id: str) -> str:
         str: string that represents the client_auth_id
     """
     course_enrollment = CourseEnrollment.objects.get(user_id=user_id, course_id=course_id)
+    anonymous_user_id = anonymous_id_for_user(course_enrollment.user, course_id)
+    anonymous_user_id_instance = AnonymousUserId.objects.get(anonymous_user_id=anonymous_user_id)
```
johanseto, line 44: Why not simply use the AnonymousUserId alone for identification? I believe it's unique per course and user. However, if some AnonymousUserId values are missing, then perhaps a combination of course_enrollment_id and anonymous_user_id could be used for tracking. That is is the reason?

andrey-canon: Same CourseEnrollment reason, if someone deletes the record we won't be able to track the user

eox_nelp/pearson_vue/utils.py
```diff
@@ -40,5 +40,7 @@ def generate_client_authorization_id(user_id: int, course_id: str) -> str:
...
 
-    return str(course_enrollment.id + CLIENT_AUTHORIZATION_ID_OFFSET)
+    return f"{course_enrollment.id + CLIENT_AUTHORIZATION_ID_OFFSET}-{anonymous_user_id_instance.id}"
```
johanseto, line 46: I think we could remove the `CLIENT_AUTHORIZATION_ID_OFFSET`.

andrey-canon: I thought you like it :)

johanseto (approved):

PR #182, opened 2024-06-20
feat: remove audited methods from api client and add them directly to…

andrey-canon:
## Description
This removes the auditable wrappers from the api_client file and move them to the corresponding pipe, this preserves the current output format however the input format has changed, this uses a dict format instead of a SOAP string.

### Before
CDD
image

EAD
image

### After
CDD
image

EAD
image

## Additional information

## Checklist for Merge

- [ ] Updated documentation
- [ ] Rebased master/main
- [ ] Squashed commits

### DIFF
--- a/eox_nelp/api_clients/pearson_rti.py
+++ b/eox_nelp/api_clients/pearson_rti.py
@@ -11,14 +11,6 @@ from django.conf import settings
 from eox_nelp.api_clients import AbstractSOAPClient
 from eox_nelp.api_clients.authenticators import PKCS12Authenticator
 
-try:
-    from eox_audit_model.decorators import audit_method
-except ImportError:
-    def audit_method(action):  # pylint: disable=unused-argument
-        """Identity audit_method"""
-        return lambda x: x
-
-
 LOGGER = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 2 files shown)


eox_nelp/api_clients/pearson_rti.py
```diff
@@ -122,41 +109,36 @@ def import_exam_authorization(self, payload):
...
+        xml_response = BeautifulSoup(
+            self.make_post("cxfws2/services/EADService", payload),
+            "xml",
+        )
+        fault = xml_response.find("soapenv:Fault")
 
+        # There are multiple kind of errors, some of them generates a soapenv:Fault section and others
+        # generates an status error section, the following condition handles the first case.
+        if fault:
             return {
```
johanseto, line 121: Maybe here we could add an exception. To try to add the eox_audit with the failure status instead of success.
 I was thinking of the audit error raising a pearsonExceptionm but it would be twice I think so... or may be not.
We have to view it after the PearsonException PR works...

johanseto (approved):
LGTM. It takes advantage that pipe step doesn't receive the credentials in kwargs.
``` python
"soapenv:Envelope": {
                "soapenv:Header": {
                    "wsse:Security": {
                        "wsse:UsernameToken"
...
}
```
I thought that we would refactor it into a step pipe like `set_pearson_credentials`.
 But in this case we have to adapt or adjust to the eox_audit model, that save the kwargs.

eox_nelp/api_clients/pearson_rti.py
```diff
@@ -122,41 +109,36 @@ def import_exam_authorization(self, payload):
...
+            }
+
+        LOGGER.error(
+            "An unexpected error has occurred trying to make a EAD request getting the following response: %s",
+            xml_response,
+        )
+
+        return {
+            "status": "unexpected error",
+            "response": xml_response,
```
johanseto, line 143: I like that any error would be returned without changes.

eox_nelp/pearson_vue/pipeline.py
```diff
@@ -224,30 +242,40 @@ def import_exam_authorization(ead_request, **kwargs):  # pylint: disable=unused-
...
+        response = api_client.import_exam_authorization(payload)
+
+        if response.get("status", "error") == "accepted":
+            return response
 
-    if response.get("status", "error") != "accepted":
+        logger.info("Import exam authorization pipeline has failed with the following response: %s", response)
         # pylint: disable=broad-exception-raised
-        raise Exception("Error trying to process import exam authorization request.")
+        raise Exception(
```
johanseto, line 274: I like that this Exception gives the _**Fail**_ status, but what do you think of the retry of the error like 4 times because it is en celery async environment.
I think that it would retry 4 times, saving the 4 Fail record. The good news is that now it is easier to distinguish than the **_success_** records.

eox_nelp/api_clients/pearson_rti.py
andrey-canon: This is the current implementation :)

eox_nelp/pearson_vue/pipeline.py
andrey-canon: Short answer out of the scope, long answer I was thinking to implement a PearsonBaseError here and limit the retrying process just to the ping pipeline since other case will need that a user o someone changes something so it's not necessary multiple attempts

johanseto: I really like that comment.

johanseto (approved):

johanseto:
Approved

PR #186, opened 2024-07-03
refactor: implement abstract backend class

andrey-canon:
## Description

Refactor that implements a backend abstract class, basivally this moves the handler error logic to a new method that could be overridden by any subclass

## Additional information

## Checklist for Merge

- [ ] Updated documentation
- [ ] Rebased master/main
- [ ] Squashed commits

### DIFF
--- a/eox_nelp/pearson_vue/rti_backend.py
+++ b/eox_nelp/pearson_vue/rti_backend.py
@@ -6,6 +6,7 @@ Classes:
     RealTimeImport: Class for managing RTI operations and executing the pipeline.
 """
 import importlib
+from abc import ABC, abstractmethod
 
 from eox_nelp.pearson_vue.exceptions import PearsonBaseError
 from eox_nelp.pearson_vue.pipeline import (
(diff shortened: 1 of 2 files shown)


eox_nelp/pearson_vue/rti_backend.py
```diff
@@ -56,20 +66,85 @@ def run_pipeline(self):
...
+        Args:
+            exception (PearsonBaseError): The exception that was raised.
+            failed_step_pipeline (str): The name of the pipeline step where the error occurred.
+        """
+
+
+class ErrorRealTimeImportHandler(AbstractBackend):
+    """Class for managing validation error pipe  executing the pipeline for data validation."""
+
+    def handle_error(self, exception, failed_step_pipeline):
```
johanseto, line 102: I think this could be removed. 
 In this case is doing nothing... or maybe we're gonna manage the error in the error pipeline?


This PR is changing because in the past the error pipeline could fail and call again the error pipeline...

andrey-canon: nop, this is necessary since the parent class is an abstract class , so all methods defined as abstract methods have to be implemented in the subclasses

johanseto (approved):
LGTM

PR #187, opened 2024-07-04
And/generate certificate based on rten request

andrey-canon:
## Description

This implements a new backend in the result notification view with new pipeline steps that allow to create a new certificate

## Testing instructions
1. set the setting `ENABLE_CERTIFICATE_PUBLISHER`, this will enable the result notification pipeline and the external certificate publisher,  its not necessary to test the external call so you can use an invalid mode, or if you want to test the external call  set `CERTIFICATE_PUBLISHER_VALID_MODES` with the desired mode
2. Make a postman call with the following data
```json
{
   "eventType":"RESULT_AVAILABLE",
   "eventTime":"?",
   "candidate":{
      "candidateName":{
         "firstName":"Maximo",
         "lastName":"Tercero"
      },
      "email":"?",
      "lastUpdate":"?",
      "vueCandidateID":"?",
      "clientCandidateID":"?"
   },
   "attempt":"$",
   "authorization":{
        "clientAuthorizationID": "1000-3000"
   },
   "exams":{
      "exam":[
         {
            "examDefinition":{
               "examSeriesCode":"?",
               "examLanguageCode":"?",
               "examName":"?",
               "deliveryModel":"?",
               "vueExamVersionId":"?",
               "vueExamRevisionID":"?",
               "clientExamVersionId":"?",
               "examForm":"?",
               "isBetaVersion":"#"
            },
            "examResult":{
               "outcomeType":"?",
               "startTime":"?",
               "timeUsed":"$",
               "passingScore":"0.8",
               "score":"0.832",
               "rawScore":"$",
               "grade":"0.5",
               "correct":"$",
               "incorrect":"$",
               "skipped":"$",
               "unscored":"$"
            },
            "noShow":"#",
            "expired":"#",
            "ndaRefused":"#",
            "deliveryEventType":"?"
         }
      ]
   },
   "vueAppointmentID":"$",
   "clientCode":"?",
   "vueCandidateID":"$",
   "clientCandidateID":"?"
}
```
3. Verify that the certificate was created in `/admin/certificates/generatedcertificate`
4. Check logs, if you didn't set `CERTIFICATE_PUBLISHER_VALID_MODES` a message with the following content must appear
`The %s certificate associated with the user <%s> and course <%s> doesn't have a valid mode and therefore its data won't be published."`

## Note
The whole integration can be tested by following this [guide](https://docs.google.com/document/d/1uzGcWyYOBqbDxzUNgtsZSk9e31P6urgzXjmM-AocZCk/edit#heading=h.7pgp0816m5td)

### DIFF
--- a/eox_nelp/pearson_vue/api/v1/views.py
+++ b/eox_nelp/pearson_vue/api/v1/views.py
@@ -34,6 +34,7 @@ from eox_nelp.pearson_vue.constants import (
     UNREVOKE_RESULT,
 )
 from eox_nelp.pearson_vue.models import PearsonRTENEvent
+from eox_nelp.pearson_vue.rti_backend import ResultNotificationBackend
 
 
 class PearsonRTENBaseView(generics.ListCreateAPIView):
@@ -130,8 +131,25 @@ class ResultNotificationView(PearsonRTENBaseView):
     This view handles the creation of Result Notification events in the Pearson RTEN system.
     The `event_type` attribute is set to "resultNotification".
     """
-
     event_type = RESULT_NOTIFICATION
+
+    def create(self, request, *args, **kwargs):
+        """
+        Execute the parent create method and allow to run the result notification pipeline.
+
+        Args:
+            request (Request): The request object containing the data.
+
+        Returns:
+            Response: Response object with status code 201 and an empty dictionary.
+        """
+        response = super().create(request, *args, **kwargs)
+
+        if response.status_code == status.HTTP_200_OK and getattr(settings, "ENABLE_CERTIFICATE_PUBLISHER", False):
+            result_notification = ResultNotificationBackend(request_data=request.data)
+            result_notification.run_pipeline()
+
+        return response
 
 
 class PlaceHoldView(PearsonRTENBaseView):
--- a/eox_nelp/pearson_vue/pipeline.py
+++ b/eox_nelp/pearson_vue/pipeline.py
@@ -32,6 +32,7 @@ from eox_nelp.pearson_vue.exceptions import (
     PearsonBaseError,
     PearsonImportError,
     PearsonKeyError,
+    PearsonTypeError,
     PearsonValidationError,
 )
 from eox_nelp.pearson_vue.utils import generate_client_authorization_id, update_xml_with_dict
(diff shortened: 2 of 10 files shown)


eox_nelp/pearson_vue/api/v1/views.py
```diff
@@ -130,9 +131,26 @@ class ResultNotificationView(PearsonRTENBaseView):
...
+            request (Request): The request object containing the data.
+
+        Returns:
+            Response: Response object with status code 201 and an empty dictionary.
+        """
+        response = super().create(request, *args, **kwargs)
+
+        if response.status_code == status.HTTP_200_OK and getattr(settings, "ENABLE_CERTIFICATE_PUBLISHER", False):
+            result_notification = ResultNotificationBackend(request_data=request.data)
+            result_notification.run_pipeline()
```
johanseto, line 150: This pipeline is run in sync. Why not async?

eox_nelp/pearson_vue/pipeline.py
```diff
@@ -496,3 +498,104 @@ def raise_audit_pearson_exception(exception_dict, failed_step_pipeline):
...
+
+    Args:
+        request_data (dict): The dictionary containing the Result Notification request data.
+        **kwargs: Additional keyword arguments.
+
+    Returns:
+        dict: A dictionary containing extracted data including exam result, client authorization ID,
+              enrollment ID, anonymous user model ID, and anonymous user ID.
+    """
+    client_authorization_id = request_data["authorization"]["clientAuthorizationID"]
```
johanseto, line 515: Are you going to handle key error with pearsonvue exceptions?? or could be other PR

eox_nelp/pearson_vue/pipeline.py
```diff
@@ -496,3 +498,104 @@ def raise_audit_pearson_exception(exception_dict, failed_step_pipeline):
...
+    """
+    Retrieves enrollment information based on the enrollment ID.
+
+    Args:
+        enrollment_id (str): The ID of the enrollment.
+        **kwargs: Additional keyword arguments.
+
+    Returns:
+        dict: A dictionary containing the enrollment object if found, otherwise an empty dictionary.
+    """
```
johanseto, line 595: Could you add the block code:
``` python3
if not enrollment:
```
In order to not be concerned it is first 
`get_enrollment_from_anonymous_user_id` or `get_enrollment_from_id`

eox_nelp/pearson_vue/api/v1/views.py
johanseto: If fails, the record would not be created in rten model

eox_nelp/pearson_vue/pipeline.py
andrey-canon: Thanks for the reminder, I will implement that in this PR

eox_nelp/pearson_vue/api/v1/views.py
andrey-canon: Let me answer with a question, Why async ?

andrey-canon: > If fails, the record would not be created in rten model

yes, probably the generics drf views implement an atomic block, but the result  will be stored if that fails due to pearson exceptions that are handled by the ErrorRealTimeImportHandler class

johanseto: > Let me answer with a question, Why async ?

Because in that case you isolate the RTEN request process from the generated certificate process.
You could avoid some 500 error codes, and maybe the client(Pearson) would think why we are getting 500...
All code related to the pipeline running in sync could make the request slower or fails and I think the Pearson Client is not aware that it is not related with the purpose of data sent...

andrey-canon: Errors should never pass silently. Python zen 
To avoid errors I can use a simple try except but I want that the client get a  500 error is the only way to know that something is happening 
request slower, yes, every line that you add in code affects the time response and when the code that you add is too much or needs a lot of time to be executed , one option is an async task, so is this that case ? have you experimented a slow response ? or have you identified a pipeline step that needs a lot of time ?

johanseto: errors would happen in the sync or async context. Anyway sentry would report both envs. So it wouldn't pass silently meanwhile sentry is working xD.
The point is who reports the error: the worker or the lms pod. And also who is being affected by that...

Anyway, if you consider is ok I could accept that.

andrey-canon: The issue isn't about what reports the errors; it's about who receives them. On one hand, we have the client using the service, and on the other, we have the Sentry service that goes unchecked until the client complains.

Finally, t main purpose of the async tasks is to execute heavy processes, not to avoid exceptions. For that, you can catch the exceptions and use a simple log.error if you want to report it to Sentry.

johanseto (approved):
Tested and working in my local env.



Also audit is managing nice the errors.

PR #190, opened 2024-07-05
feat: add candidate field to RTEN model

andrey-canon:
## Description

This adds a new User field to the RTEN model so that the record is associated with a User record in the database. This includes the required view modifications 

## Testing instructions
1. Run migrations `./manage.py lms migrate eox_nelp`
2. set the setting `ENABLE_CERTIFICATE_PUBLISHER`, this will enable the result notification pipeline and the external certificate publisher,  its not necessary to test the external call so you can use an invalid mode, or if you want to test the external call  set `CERTIFICATE_PUBLISHER_VALID_MODES` with the desired mode
3. Make a postman call with the following data
```json
{
   "eventType":"RESULT_AVAILABLE",
   "eventTime":"?",
   "candidate":{
      "candidateName":{
         "firstName":"Maximo",
         "lastName":"Tercero"
      },
      "email":"?",
      "lastUpdate":"?",
      "vueCandidateID":"?",
      "clientCandidateID":"?"
   },
   "attempt":"$",
   "exams":{
      "exam":[
         {
            "examDefinition":{
               "examSeriesCode":"?",
               "examLanguageCode":"?",
               "examName":"?",
               "deliveryModel":"?",
               "vueExamVersionId":"?",
               "vueExamRevisionID":"?",
               "clientExamVersionId":"?",
               "examForm":"?",
               "isBetaVersion":"#"
            },
            "examResult":{
               "outcomeType":"?",
               "startTime":"?",
               "timeUsed":"$",
               "passingScore":"0.8",
               "score":"0.832",
               "rawScore":"$",
               "grade":"0.5",
               "correct":"$",
               "incorrect":"$",
               "skipped":"$",
               "unscored":"$"
            },
            "noShow":"#",
            "expired":"#",
            "ndaRefused":"#",
            "deliveryEventType":"?"
         }
      ]
   },
   "vueAppointmentID":"$",
   "clientCode":"?",
   "vueCandidateID":"$",
   "clientCandidateID":"?" /// Complete this
}
```
4. Check in admin panel
### Before
image

### After
image

## Additional information

## Checklist for Merge

- [ ] Updated documentation
- [ ] Rebased master/main
- [ ] Squashed commits


eox_nelp/pearson_vue/models.py
```diff
@@ -47,3 +49,4 @@ class PearsonRTENEvent(models.Model):
     content = models.JSONField()
     created_at = models.DateTimeField(auto_now_add=True)
     event_type = models.CharField(max_length=20, choices=EVENT_TYPE_CHOICES)
+    candidate = models.ForeignKey(User, null=True, on_delete=models.SET_NULL)
```
johanseto, line 52: Do you think we could use `blank=True`, in order to don't be mandatory in forms or serializer.?

eox_nelp/pearson_vue/api/v1/views.py
```diff
@@ -122,6 +123,24 @@ def create(self, request, *args, **kwargs):
...
+        Retrieves the candidate user based on the client candidate ID provided in the request data.
+
+        This method extracts the `clientCandidateID` from the request data, removes the "NELC" prefix,
+        and attempts to retrieve the corresponding user from the `AnonymousUserId` model. If the candidate
+        ID does not exist, it returns `None`.
+
+        Returns:
+            user (object or None): The user object associated with the anonymous user ID, or `None` if not found.
+        """
+        anonymous_user_id = self.request.data.get("clientCandidateID", "").replace("NELC", "")
```
johanseto, line 137: Ideally, we'd avoid modifying clientCandidateId(product requirement?) due to its impact on product functionality and user experience. Any changes would necessitate significant code updates related to anonymous IDs.

eox_nelp/pearson_vue/models.py
andrey-canon: I wanted to do this as mandatory as possible since the idea is that every record has a user however we don't know if the client will send the whole data in the example format and secondly the migration process complains about previous records when the null=True field is not present, that's why I added the null field and I don't think we should include the blank field

johanseto (approved):
Tested and working in my local env.

PR #192, opened 2024-07-08
feat: add audit for succesfull imports

johanseto:
## Description
add audit to pipelines

### After
image

## Additional information

## Checklist for Merge

- [ ] Updated documentation
- [ ] Rebased master/main
- [ ] Squashed commits

### DIFF
--- a/eox_nelp/pearson_vue/pipeline.py
+++ b/eox_nelp/pearson_vue/pipeline.py
@@ -227,7 +227,7 @@ def import_candidate_demographics(cdd_request, **kwargs):  # pylint: disable=unu
     response = api_client.import_candidate_demographics(payload)
 
     if response.get("status", "error") == "accepted":
-        return response
+        return {"cdd_import": response}
 
     raise PearsonImportError(
         exception_reason=f"Import candidate demographics pipeline has failed with the following response: {response}",
@@ -276,7 +276,7 @@ def import_exam_authorization(ead_request, **kwargs):  # pylint: disable=unused-
     response = api_client.import_exam_authorization(payload)
 
     if response.get("status", "error") == "accepted":
-        return response
+        return {"ead_import": response}
 
     raise PearsonImportError(
         exception_reason=f"Import exam authorization pipeline has failed with the following response: {response}",
--- a/eox_nelp/pearson_vue/rti_backend.py
+++ b/eox_nelp/pearson_vue/rti_backend.py
@@ -53,6 +53,7 @@ class AbstractBackend(ABC):
         get_pipeline(): Returns the pipeline, which is a list of functions to be executed (abstract method).
         handle_error(exception: Exception, failed_step_pipeline: str): Handles errors during pipeline execution.
     """
+    use_audit_backend = True
 
     def __init__(self, **kwargs):
         """
@@ -63,6 +64,7 @@ class AbstractBackend(ABC):
         """
         self.backend_data = kwargs.copy()
 
+    @audit_backend
     def run_pipeline(self):
         """
         Executes the pipeline by iterating through the pipeline functions.
@@ -141,9 +143,7 @@ class RealTimeImport(AbstractBackend):
         run_pipeline(): Executes the RTI pipeline by iterating through the pipeline functions.
         get_pipeline(): Returns the RTI pipeline, which is a list of functions to be executed.
     """
-    @audit_backend
-    def run_pipeline(self):
-        return super().run_pipeline()
+    use_audit_backend = False
 
     def handle_error(self, exception, failed_step_pipeline):
         """
--- a/eox_nelp/pearson_vue/rti_backend.py
+++ b/eox_nelp/pearson_vue/rti_backend.py
@@ -81,6 +81,7 @@ class AbstractBackend(ABC):
             try:
                 result = func(**self.backend_data) or {}
             except PearsonBaseError as pearson_error:
+                self.backend_data["catched_pearson_error"] = True
                 self.handle_error(pearson_error, func.__name__)
                 break
 
@@ -113,6 +114,7 @@ class AbstractBackend(ABC):
 
 class ErrorRealTimeImportHandler(AbstractBackend):
     """Class for managing validation error pipe  executing the pipeline for data validation."""
+    use_audit_backend = False
 
     def handle_error(self, exception, failed_step_pipeline):
         """
@@ -143,8 +145,6 @@ class RealTimeImport(AbstractBackend):
         run_pipeline(): Executes the RTI pipeline by iterating through the pipeline functions.
         get_pipeline(): Returns the RTI pipeline, which is a list of functions to be executed.
     """
-    use_audit_backend = False
-
     def handle_error(self, exception, failed_step_pipeline):
         """
         Handles errors during pipeline execution.
--- a/eox_nelp/pearson_vue/rti_backend.py
+++ b/eox_nelp/pearson_vue/rti_backend.py
@@ -16,6 +16,7 @@ Classes:
 import importlib
 from abc import ABC, abstractmethod
 
+from eox_nelp.pearson_vue.decorators import audit_backend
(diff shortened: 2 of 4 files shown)


andrey-canon (commented):
why you modified  current pipeline steps instead of adding a general pipeline step that allows to audit the whole process ?

johanseto:
> why you modified current pipeline steps instead of adding a general pipeline step that allows to audit the whole process ?

I think the scope was only for the successful import, do you think I would audit all the pipeline execution?

eox_nelp/pearson_vue/pipeline.py
```diff
@@ -613,3 +614,43 @@ def get_enrollment_from_id(enrollment_id, enrollment=None, **kwargs):  # pylint:
             pass
 
     return {}
+
+
+def save_audit_pipeline(pipeline_backend, last_step, cdd_import=None, ead_import=None, **kwargs):
```
andrey-canon, line 619: As you mentioned this is not a pipeline step, so I don't understand why you included this in the **pipeline** file, anyway I wondering why you split the logic into this method and audit pipeline decorator

eox_nelp/pearson_vue/pipeline.py
```diff
@@ -613,3 +614,43 @@ def get_enrollment_from_id(enrollment_id, enrollment=None, **kwargs):  # pylint:
...
+    """This is not a pipe step, could be but its no the intention because inside a pipeline could
+    be more than one import. so this is a helper for pipelines.
+
+    Args:
+        pipeline_backend (str):  pipeline class name
+        last_step (str): last function executed
+        cdd_import (dict, optional): Defaults to None.
+        ead_import (dict, optional):  Defaults to None.
```
andrey-canon, line 627: Why ? this implementation is too specific, you are assuming in advanced that the backend will implement cdd and ead pipelines steps and that is not true, probably your answer will include that there are some  default values  but again this code was written to work with something specific instead of a general solution

eox_nelp/pearson_vue/pipeline.py
```diff
@@ -613,3 +614,43 @@ def get_enrollment_from_id(enrollment_id, enrollment=None, **kwargs):  # pylint:
...
+    """
+
+    @audit_method(action=f"Pearson Vue Pipeline Execution: {pipeline_backend}")
+    @rename_function(name=f"audit_pipeline_{camel_to_snake(pipeline_backend)}")
+    def executed_pipeline(
+        pipeline_backend,
+        last_step,
+        cdd_import,
+        ead_import,
+        **kwargs
```
andrey-canon, line 637:
```suggestion
    def executed_pipeline(
        backend_name,
        last_pipeline_step,
        backend_data,
```

eox_nelp/pearson_vue/pipeline.py
```diff
@@ -613,3 +614,43 @@ def get_enrollment_from_id(enrollment_id, enrollment=None, **kwargs):  # pylint:
...
+        logger.info(
+            "Pipeline %s executed until last step: %s",
+            pipeline_backend,
+            last_step,
+        )
+        logger.info(
+            "Succesfull imports. cdd_import: %s. ead_import: %s",
+            cdd_import,
+            ead_import,
+        )
```
andrey-canon, line 648:
```suggestion
```
Since this is not a general implementation

eox_nelp/pearson_vue/rti_backend.py
```diff
@@ -31,11 +31,36 @@
...
+    Returns:
+        A wrapper function that executes the decorated method with a try-finally block.
+    """
+    def wrapper(self, *args, **kwargs):
+        try:
+            # Call the original function with self, args, and kwargs
+            return func(self, *args, **kwargs)
+        finally:
+            if self.use_audit_pipeline and not self.backend_data.get("catched_pearson_error"):
+                save_audit_pipeline(
```
andrey-canon, line 55: This is just a simple log, why you didn't include the logic here ?

eox_nelp/pearson_vue/rti_backend.py
```diff
@@ -31,11 +31,36 @@
...
+                    pipeline_backend=self.__class__.__name__,
+                    last_step=self.get_pipeline()[self.backend_data.get("pipeline_index", 0)].__name__,
+                    **self.backend_data
```
andrey-canon, line 58:
```suggestion
                   backend_data=self.backend_data
```

eox_nelp/pearson_vue/rti_backend.py
```diff
@@ -53,15 +78,17 @@ class AbstractBackend(ABC):
         handle_error(exception: Exception, failed_step_pipeline: str): Handles errors during pipeline execution.
     """
 
-    def __init__(self, **kwargs):
+    def __init__(self, use_audit_pipeline=True, **kwargs):
```
andrey-canon, line 81: The advantage of this is that you can change the backend behavior based on the initialization, but  do we need that?, it's not better stable backends whose execution or behavior is always the same ?

```suggestion
    use_audit_pipeline=True
    
    def __init__(self, **kwargs):
```

eox_nelp/pearson_vue/rti_backend.py
```diff
@@ -31,11 +31,36 @@
     handle_course_completion_status,
     import_candidate_demographics,
     import_exam_authorization,
+    save_audit_pipeline,
     validate_cdd_request,
     validate_ead_request,
 )
 
 
+def audit_pipeline(func):
```
andrey-canon, line 40: could you please move this to a decorators.py file ?

andrey-canon (changes requested):

eox_nelp/pearson_vue/rti_backend.py
johanseto: Yes, but then I must apply the decorator to the specific pipeline backend. Otherwise I would get something like this.
https://edunext.stage.nelp.gov.sa/admin/eox_audit_model/auditmodel/32026/change/

johanseto: You can see the error pipeline is also audited.

andrey-canon: what happen if you override that in the subclass ?

eox_nelp/pearson_vue/pipeline.py
```diff
@@ -227,7 +227,7 @@ def import_candidate_demographics(cdd_request, **kwargs):  # pylint: disable=unu
     response = api_client.import_candidate_demographics(payload)
 
     if response.get("status", "error") == "accepted":
-        return response
+        return {"cdd_import": response}
```
andrey-canon, line 230: do you think we will need this ?

eox_nelp/pearson_vue/pipeline.py
```diff
@@ -276,7 +276,7 @@ def import_exam_authorization(ead_request, **kwargs):  # pylint: disable=unused-
     response = api_client.import_exam_authorization(payload)
 
     if response.get("status", "error") == "accepted":
-        return response
+        return {"ead_import": response}
```
andrey-canon, line 279: Same comments as above

andrey-canon (commented):
why did you remove some things like the  `use_audit_pipeline` and the `catched_pearson_error`?

johanseto:
> why did you remove some things like the `use_audit_pipeline` and the `catched_pearson_error`?

The use audit because don't you recommend using the override in the `run_pipeline` method?

andrey-canon:
> > why did you remove some things like the `use_audit_pipeline` and the `catched_pearson_error`?
> 
> The use audit because don't you recommend using the override in the `run_pipeline` method?

Nop, there was a misunderstanding, my previous recommendation was to test  use_audit_pipeline as attribute class and then override that in the subclass

```python
class AbstractBackend(ABC):
    use_audit_pipeline=True

class ErrorRealTimeImportHandler(AbstractBackend):
    use_audit_pipeline=False
```

andrey-canon (approved):

PR #199, opened 2024-07-15
refactor: replace generic list create api view with generic viewset a…

andrey-canon:
## Description

This doesn't modify anything; it simply changes the current implementation to use a viewset implementation and updates the create method to utilize the super method instead of duplicating it.

### DIFF
--- a/eox_nelp/pearson_vue/api/v1/views.py
+++ b/eox_nelp/pearson_vue/api/v1/views.py
@@ -18,9 +18,11 @@ from django.conf import settings
 from django.http import Http404
 from edx_rest_framework_extensions.auth.jwt.authentication import JwtAuthentication
 from eox_core.edxapp_wrapper.bearer_authentication import BearerAuthentication
-from rest_framework import generics, status
+from rest_framework import status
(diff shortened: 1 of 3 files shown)


eox_nelp/pearson_vue/api/v1/views.py
```diff
@@ -124,11 +126,12 @@ def create(self, request, *args, **kwargs):
...
             Response: Response object with status code 201 and an empty dictionary.
         """
-        serializer = self.get_serializer(data=request.data)
-        serializer.is_valid(raise_exception=True)
-        self.perform_create(serializer)
-        headers = self.get_success_headers(serializer.data)
-        return Response({}, status=status.HTTP_200_OK, headers=headers)
+        response = super().create(request, *args, **kwargs)
+
+        if response.status_code == status.HTTP_201_CREATED:
```
johanseto, line 131: This is so crazy, chaging normal status codes. But we depends on client logic 201 ->200

johanseto (approved):
Wroking as expected. The only thing is the is necessary eh fix of https://github.com/eduNEXT/eox-nelp/pull/201/commits

PR #201, opened 2024-07-15
fix: turn off audit feature

andrey-canon:
## Description

Deactivate audit process, since the backend has data that is not serializable 

### Before
image

### After
image

## Additional information

## Checklist for Merge

- [ ] Updated documentation
- [ ] Rebased master/main
- [ ] Squashed commits

### DIFF
--- a/eox_nelp/pearson_vue/rti_backend.py
+++ b/eox_nelp/pearson_vue/rti_backend.py
@@ -220,6 +220,7 @@ class ResultNotificationBackend(RealTimeImport):
     for handling result notifications, including extracting result notification data,
     retrieving enrollment information, and generating external certificates.
     """
+    use_audit_backend = False
 
     def get_pipeline(self):
         """


johanseto (approved):

eox_nelp/pearson_vue/rti_backend.py
```diff
@@ -220,6 +220,7 @@ class ResultNotificationBackend(RealTimeImport):
     for handling result notifications, including extracting result notification data,
     retrieving enrollment information, and generating external certificates.
     """
+    use_audit_backend = False
```
johanseto, line 223: Do you thinks is not important to audit  ResultNotificationBackend
image

andrey-canon: I think that is important but it's not possible, anyway, the certificate has its own audited method so we won't be able to audit the whole pipeline but we will get the same current behavior for external certificates

PR #203, opened 2024-07-16
feat: add filter backends and limit the list view to super or staff u…

andrey-canon:
## Description

1. Added SearchFilter with search_fields ["candidate__username", "course__id"] to enable search functionality.
2. Updated the list method of the viewset to restrict access to staff or superusers only.

##### Change Objective:
Enhance endpoint security and functionality by limiting access to authorized users.
Facilitate event searches by user and course using the SearchFilter.

## Testing instructions
1. Make a request to `/eox-nelp/api/pearson_vue/v1/resultNotification/` with a staff or super user, expected behavior => list of records returned 
2. Make a request to `/eox-nelp/api/pearson_vue/v1/resultNotification/` with a non staff , expected behavior => 404
3. Make a request with the search query param `/eox-nelp/api/pearson_vue/v1/resultNotification/?search=course-v1:edx%2BCS101%2B2023_t4`  or `/eox-nelp/api/pearson_vue/v1/resultNotification/?search=<username>` expected behavior => filtered list of records

### DIFF
--- a/eox_nelp/pearson_vue/api/v1/views.py
+++ b/eox_nelp/pearson_vue/api/v1/views.py
@@ -19,6 +19,7 @@ from django.http import Http404
 from edx_rest_framework_extensions.auth.jwt.authentication import JwtAuthentication
 from eox_core.edxapp_wrapper.bearer_authentication import BearerAuthentication
 from rest_framework import status
+from rest_framework.filters import SearchFilter
 from rest_framework.mixins import CreateModelMixin, ListModelMixin, RetrieveModelMixin
 from rest_framework.permissions import IsAuthenticated
 from rest_framework.response import Response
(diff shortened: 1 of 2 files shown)


eox_nelp/pearson_vue/api/v1/views.py
```diff
@@ -133,6 +136,30 @@ def create(self, request, *args, **kwargs):
...
+            *args: Additional positional arguments.
+            **kwargs: Additional keyword arguments.
+
+        Returns:
+            Response: The response object containing the list of objects if the user has the necessary permissions.
+
+        Raises:
+            Http404: If the user does not have the necessary permissions.
+        """
+        if request.user.is_superuser or request.user.is_staff:
```
johanseto, line 158: So this force that the User used for the application to get the token, needs to be staff?
I think that only creating the application was enough to get a token and use the API.

andrey-canon: if you are not staff you shouldn't be able to get data from another users

PR #205, opened 2024-07-17
feat: userSignUpSource create by course_enrollment

johanseto:
## Description

feat: userSignUpSource created by course_enrollment

Configure signal after a courseEnrollment save to create the desired UserSignUpSource record based on the course_enrollment using the org course.

### After
1. Go to UserSignUpSource in admin and select a user.

2. Go to course_enrollment in admin and select a course_enrollment related to the user and a course with an org-related site that doesn't have a UserSignUpSource record. Save the course_enrollment.

3. Check logs or check that the UserSignUpSource record was created.

## Checklist for Merge

- [ ] Updated documentation
- [ ] Rebased master/main
- [ ] Squashed commits

### DIFF
--- a/eox_nelp/signals/receivers.py
+++ b/eox_nelp/signals/receivers.py
@@ -19,6 +19,8 @@ import logging
 from django.conf import settings
 from django.contrib.auth import get_user_model
 from eox_core.edxapp_wrapper.grades import get_course_grade_factory
+from eox_core.edxapp_wrapper.users import get_user_signup_source
+from eox_tenant.tenant_wise.proxies import TenantSiteConfigProxy
 from eventtracking import tracker
 from openedx_events.learning.data import CertificateData, CourseData, UserData, UserPersonalData
 
@@ -35,6 +37,7 @@ from eox_nelp.signals.tasks import (
 from eox_nelp.signals.utils import _generate_external_certificate_data
 
 User = get_user_model()
+UserSignupSource = get_user_signup_source()
 LOGGER = logging.getLogger(__name__)
 CourseGradeFactory = get_course_grade_factory()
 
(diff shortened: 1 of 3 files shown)


eox_nelp/signals/receivers.py
```diff
@@ -227,6 +230,38 @@ def enrollment_publisher(instance, **kwargs):  # pylint: disable=unused-argument
...
+    """
+    Receiver that is connected to the course enrollment post_save signal. This will generate a
+    UserSignupSource record if possilbe
+    based on the `SITE_NAME` tenant settings.
+    If there is not SITE_NAME configurated this would be skipped with a log.
+
+    Args:
+        instance<CourseEnrollment>: This an instance of the model CourseEnrollment.
+    """
+    site_name = TenantSiteConfigProxy.get_value_for_org(instance.course_id.org, "SITE_NAME")
```
andrey-canon, line 243:
```suggestion
    site_name = getattr(settings, "SITE_NAME", None)
```
I'm just wondering if this is not the same ? and I worry about the tenants that have multiple orgs

johanseto: @andrey-canon Because you could save a courseEnrollment of lms org using the edunext tenant. So the signUpSource would be attached to the org of the courseEnrollment and not the current tenant request. Also in async could help to don't think in the async tenant.

johanseto: According to tenants having multiple orgs. The think is fetching inverse. The courseEnrollment has a course eg
`course-v1:BRAGI+cd101+2023t1`. That course would match org BRAGI. Then the method would search for a tenantConfig that matches that organization in org_filter(organizations).
After that in the TenanConfig matched, the `SITE_NAME` is retrieved.
The problem could be with multiple tenant_config that have the same org. In this case,
the first one would be selected, according to the following [snippet](https://github.com/eduNEXT/eox-tenant/blob/c4b4b6c716b6b03763c076156b3595ac2c324df2/eox_tenant/models.py#L218-L226).

andrey-canon (approved):

PR #213, opened 2024-08-05
refactor: extract completion logic from pipe to implement it into rec…

andrey-canon:
## Description

This move the completion logic from the pipeline step to the receiver, this also adds some logs to  monitor which actions is triggered the pearson task 

## Additional information

## Checklist for Merge

- [ ] Updated documentation
- [ ] Rebased master/main
- [ ] Squashed commits


eox_nelp/signals/receivers.py
```diff
@@ -393,18 +393,26 @@ def pearson_vue_course_completion_handler(instance, **kwargs):  # pylint: disabl
...
+                user_id=instance.user_id,
+                exam_id=str(instance.context_key),
+                action_name="rti",
+            )
+        else:
+            real_time_import_task.delay(
+                user_id=instance.user_id,
+                course_id=str(instance.context_key),
+            )
+
```
johanseto, line 415:
```suggestion
    if (not is_complete) or graded:
        return
 
    LOGGER.info(
        "Initializing rti task for the user %s, action triggered by course completion status",
        instance.user_id,
    )

    if getattr(settings, "USE_PEARSON_ENGINE_SERVICE", False):
        real_time_import_task_v2.delay(
            user_id=instance.user_id,
            exam_id=str(instance.context_key),
            action_name="rti",
        )
    else:
        real_time_import_task.delay(
            user_id=instance.user_id,
            course_id=str(instance.context_key),
        )

```

johanseto (commented):
I made a logic suggestion. I think test would pass with that change.

eox_nelp/signals/receivers.py
andrey-canon: very similar to my first approach `if  graded or not is_complete:` but I decided to leave as similar as possible in order to make easier the review, anyway I will update that

johanseto (approved):

PR #216, opened 2024-09-02
fix: not start rti tasks if not course-config

johanseto:
# Description

Check the existence of course-config for the processed course_id.
If there is not course-config (`PEARSON_RTI_COURSES_DATA`) for the desired course,
RTI requests would not be sent.

Refactored a little the tests.

## Additional information

**[Jira story](https://edunext.atlassian.net/browse/FUTUREX-877?atlOrigin=eyJpIjoiMjNiMzZmNmU3ZWU4NDhmNzg0N2EzNTM4ZmNkZDg3MTkiLCJwIjoiaiJ9)**: 

## Checklist for Merge

- [ ] Updated documentation
- [ ] Rebased master/main
- [ ] Squashed commits

### DIFF
--- a/eox_nelp/signals/receivers.py
+++ b/eox_nelp/signals/receivers.py
@@ -386,15 +386,13 @@ def pearson_vue_course_completion_handler(instance, **kwargs):  # pylint: disabl
     the pearson vue real_time_import_task , that basically runs the RTI pipeline
     on the completion logic.
     This sends only the user_id and course_id kwargs.
-    user_id -> instance.user_id
-    course_key -> instance.context_key
 
     Arguments:
         instance<Blockcompletion>: Instance of BlockCompletion model.
     """
-    if not getattr(settings, "PEARSON_RTI_ACTIVATE_COMPLETION_GATE", False) or not getattr(
-        settings, "PEARSON_RTI_COURSES_DATA", {}
-    ).get(str(instance.context_key)):
+    if not getattr(settings, "PEARSON_RTI_ACTIVATE_COMPLETION_GATE", False) or not str(instance.context_key) in getattr(
+        settings, "PEARSON_ENGINE_COURSES_ENABLED", []
+    ):
         return
 
     is_complete, graded = get_completed_and_graded(user_id=instance.user_id, course_id=str(instance.context_key))
@@ -430,9 +428,9 @@ def pearson_vue_course_passed_handler(user, course_id, **kwargs):  # pylint: dis
         user <User>: Instance of auth user model.
         course_id <CourseLocator>: Course locator.
     """
-    if not getattr(settings, "PEARSON_RTI_ACTIVATE_GRADED_GATE", False) or not getattr(
-        settings, "PEARSON_RTI_COURSES_DATA", {}
-    ).get(str(course_id)):
+    if not getattr(settings, "PEARSON_RTI_ACTIVATE_GRADED_GATE", False) or not str(course_id) in getattr(
+        settings, "PEARSON_ENGINE_COURSES_ENABLED", []
+    ):
         return
 
     LOGGER.info(
--- a/eox_nelp/signals/receivers.py
+++ b/eox_nelp/signals/receivers.py
@@ -392,9 +392,7 @@ def pearson_vue_course_completion_handler(instance, **kwargs):  # pylint: disabl
     Arguments:
         instance<Blockcompletion>: Instance of BlockCompletion model.
     """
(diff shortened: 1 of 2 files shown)


eox_nelp/signals/receivers.py
```diff
@@ -386,11 +386,15 @@ def pearson_vue_course_completion_handler(instance, **kwargs):  # pylint: disabl
...
     This sends only the user_id and course_id kwargs.
+    user_id -> instance.user_id
+    course_key -> instance.context_key
 
     Arguments:
         instance<Blockcompletion>: Instance of BlockCompletion model.
     """
-    if not getattr(settings, "PEARSON_RTI_ACTIVATE_COMPLETION_GATE", False):
+    if not getattr(settings, "PEARSON_RTI_ACTIVATE_COMPLETION_GATE", False) or not getattr(
+        settings, "PEARSON_RTI_COURSES_DATA", {}
```
andrey-canon, line 396: please consider to update the setting name and type

johanseto: The question is: is someone using the past integration? The idea was to keep compatibility with both integrations.

Anyway why fo you say the type? to change the dict to a list or something like that? I also like the dict approach:
``` json
{
"PEARSON_RTI_COURSES_DATA": {
"course-v1:one": true,
"course-v1:two": true,
"course-v1:three": false

...
}
}
```

andrey-canon: >The question is: is someone using the past integration? The idea was to keep compatibility with both integrations.

If the setting name has been changed, it doesn’t mean the previous implementation is unavailable. It might require an extra configuration step, which could seem redundant, but it remains compatible.

> Anyway why fo you say the type? to change the dict to a list or something like that? I also like the dict approach:

You mentioned compatibility, but the dictionary approach is indeed completely new; the current implementation does not support that structure. However, if you prefer the dictionary over a simple list, that’s fine. Just update the setting name and adjust the logic to check for the value instead of the key.

johanseto: Actually the v0 version is also compatible looking the test:
https://github.com/eduNEXT/eox-nelp/blob/7785b9bbc78aeabc5ffefd60ef62eb485dd91f77/eox_nelp/signals/tests/test_receivers.py#L703-L710

Is like 
``` json
{
"PEARSON_RTI_COURSES_DATA": {
  "course-v1:one": { 
      "exam_authorization_count": 3, 
      "exam_series_code": "OTT" 
  },
  "course-v1:two": { 
      "exam_authorization_count": 23, 
      "exam_series_code": "OTT" 
  },
  "course-v1:three": { 
      "exam_authorization_count": 13, 
      "exam_series_code": "OTT" 
  }

...
}
```
so each course if the value is python truthy like a defined dict: it would pass.

So what's the purpose of creating another setting for courses, if the current could work in both versions

So in both examples I am checking the value of the key:
with the dict get
https://github.com/eduNEXT/eox-nelp/blob/7785b9bbc78aeabc5ffefd60ef62eb485dd91f77/eox_nelp/signals/receivers.py#L395-L397

andrey-canon: >So what's the purpose of creating another setting for courses, if the current could work in both versions

basically the name of the settings doesn't reflect the purpose of the setting

johanseto: I added in this 2 commits.
609622da0132b153cac73f2cb883aa330c8de03e
[f72c455](https://github.com/eduNEXT/eox-nelp/pull/216/commits/f72c4552434bdccaa863c8a802f8bf997dd17b6c)
Any way, I don't like it. Now, we have to manage two settings, one for v1 and the other for the engine.
Also, we have to synchronize those settings via query or something like that to configure each course with the Pearson version to use.

andrey-canon: 1. You don't have to manage two settings previous implementation is deprecated  and you shouldn't use it
2. You didn't have to add the second condition that implementation fails if that setting is not present 
3. The previous implementation is deprecated therefore you shouldn't modify that at all
4. You don't need any query, it's faster if you do that manually there are only two courses

johanseto:
@andrey-canon COuld you check if you agree with [f125f1e](https://github.com/eduNEXT/eox-nelp/pull/216/commits/f125f1e9c0a2eee76243680a6828b93ccea41827).

The setting now would have this shape. And I have to update manually the 2 courses or tenant_config using the PVengine integration.
``` json
"PEARSON_ENGINE_COURSES_ENABLED" : ["course-v1:one", "course-v1:two"]
```
PD: Tests are falling due granular fine token permission to clone custom_reg_form.

andrey-canon (approved):

PR #222, opened 2024-10-09
feat: replace username with extrainfo national_id

andrey-canon:
## Description

This replaces the username references with the  extrainfo national id field 

## Additional information

## Checklist for Merge

- [ ] Updated documentation
- [ ] Rebased master/main
- [ ] Squashed commits

### DIFF
--- a/eox_nelp/signals/receivers.py
+++ b/eox_nelp/signals/receivers.py
@@ -356,7 +356,7 @@ def mt_course_passed_handler(user, course_id, **kwargs):  # pylint: disable=unus
         return
 
     extra_info = getattr(user, "extrainfo", None)
-    national_id = extra_info.national_id if extra_info else user.username
+    national_id = extra_info.national_id if extra_info and extra_info.national_id else user.username
 
     update_mt_training_stage.delay(
         course_id=str(course_id),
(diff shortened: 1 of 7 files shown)


eox_nelp/signals/tests/test_receivers.py
```diff
@@ -14,7 +14,7 @@
 from ddt import data, ddt
 from django.conf import settings
 from django.contrib.auth import get_user_model
-from django.test import override_settings
+from django.test import TestCase, override_settings
```
johanseto, line 17: Why you need to use Django test case?

eox_nelp/signals/receivers.py
```diff
@@ -355,9 +355,12 @@ def mt_course_passed_handler(user, course_id, **kwargs):  # pylint: disable=unus
     if not getattr(settings, "ACTIVATE_MT_TRAINING_STAGE", False):
         return
 
+    extra_info = getattr(user, "extrainfo", None)
+    national_id = extra_info.national_id if extra_info else user.username
```
johanseto, line 359: What about if national_id is "". that would be sent?. Is weird to found same usernames. But this this field of extra_info seems allow blank

eox_nelp/signals/tests/test_receivers.py
andrey-canon: Becasue unittest.TestCase doesn't work with reverse relations, so things like user.extrainfo fails

johanseto (approved):

PR #230, opened 2024-10-31
Jlc/extende usersupport admin

johanseto:
## Description

feat: add nelpUser admin

This admin user adds:
- national_id from extra_info instance in list_display
- search by national_id

### After

- Check lms user admin has national_id field
- search by national_id in user admin

[Screencast from 31-10-24 16:59:09.webm](https://github.com/user-attachments/assets/c467dba9-48c1-4b52-82ca-ec6d00c6f4db)

- check studio is working and user admin is not modified

Screenshot from 2024-11-01 11-23-00

## Additional information
**[jira story](https://edunext.atlassian.net/browse/FUTUREX-1055?atlOrigin=eyJpIjoiMWNjMDU1Y2MzMGFlNDUxYmEwNzRmNWYzNjYzMjQ0ZGMiLCJwIjoiaiJ9)**

## Checklist for Merge

- [ ] Updated documentation
- [ ] Rebased master/main
- [ ] Squashed commits

### DIFF
--- a/eox_nelp/admin/user.py
+++ b/eox_nelp/admin/user.py
@@ -29,7 +29,6 @@ class NelpUserAdmin(SupportUserAdmin):
     list_display = SupportUserAdmin.list_display[:2] + ('user_national_id',) + SupportUserAdmin.list_display[2:]
     search_fields = SupportUserAdmin.search_fields + ('extrainfo__national_id',)
     inlines = SupportUserAdmin.inlines + (UserExtraInfoInline,)
-    readonly_fields = SupportUserAdmin.readonly_fields + ('user_national_id',)
 
     def user_national_id(self, instance):
         """Return national_id associated with the user extra_info instance."""
--- a/eox_nelp/admin/user.py
+++ b/eox_nelp/admin/user.py
@@ -1,0 +1,26 @@
+from django.contrib.auth import get_user_model
+from eox_support.admin.user import SupportUserAdmin
+
+from eox_nelp.admin.register_admin_model import register_admin_model as register
+
+User = get_user_model()
+
+
+class NelpUserAdmin(SupportUserAdmin):
+    """EoxNelp User admin class."""
+
+    list_display = ('user_national_id',) + SupportUserAdmin.list_display
+    search_fields = SupportUserAdmin.search_fields + ('extrainfo__national_id',)
+    fieldsets = SupportUserAdmin.fieldsets + (
+        ('Extra info Fields', {'fields': ('user_national_id',)}),
+    )
+    readonly_fields = SupportUserAdmin.readonly_fields + ('user_national_id',)
+    def user_national_id(self, instance):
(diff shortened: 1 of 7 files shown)


eox_nelp/admin/user.py
```diff
@@ -0,0 +1,42 @@
...
+
+User = get_user_model()
+
+
+class NelpUserAdmin(SupportUserAdmin):
+    """EoxNelp User admin class."""
+    list_display = SupportUserAdmin.list_display[:2] + ('user_national_id',) + SupportUserAdmin.list_display[2:]
+    search_fields = SupportUserAdmin.search_fields + ('extrainfo__national_id',)
+    fieldsets = SupportUserAdmin.fieldsets + (
+        ('Extra info Fields', {'fields': ('user_national_id',)}),
```
andrey-canon, line 29: could you include this in the personal info section ?

eox_nelp/admin/user.py
andrey-canon, line 1: I think you can simplify this file, basically the new model only has effects if eox-support is installed 
so if you move this condition `if find_spec('eox_support') and 'eox_support.apps.EoxSupportConfig' in settings.INSTALLED_APPS:` to a higher level ( eox_vnelp/admin/__init__.py) you can remove all those repeated validations from this file.

Just one question, why you didn't include eox-support as requierement?

andrey-canon (changes requested):

eox_nelp/admin/user.py
johanseto: This is basically because the admin of eox_support uses a lot of edx_app wrappers. So, this can started to bother with this kind of error in the test environment.

Screenshot from 2024-10-31 18-22-25

eox_nelp/admin/user.py
johanseto: Why you dont want to split to extra info fields. Here could be first_arabic_name, or is_phone_validated. 
That fields belongs to extra_info model. So they should go to that table or admin model to check

andrey-canon: This kind of classification just makes sense for us, for someone who doesn't know about tables or models this could be  more confusing, on the other hand, if the purpose is to facilitate  the navigation to extra info table, don't add a national id, add a link to the instance, anyway, if you consider that is too hard or that you have to copy the whole Personal info, to achieve that, please keep this way

johanseto: [Screencast from 05-11-24 11:48:23.webm](https://github.com/user-attachments/assets/6a4f852d-ca19-4941-bff8-5358e5c7aef5)
I like the inline option =)

eox_nelp/admin/user.py
johanseto: Also moved to init file the if condition.
[606b4eb](https://github.com/eduNEXT/eox-nelp/pull/230/commits/606b4eb9144893ad52f724ab9d8f1324a17f877f)

andrey-canon (approved):
LGTM

PR #233, opened 2024-11-14
feat: create a new pipeline step that validates the national id base …

andrey-canon:
## Description

This pull request introduces the validate_national_id_and_associate_user pipeline method, designed to improve user association logic in SAML authentication flows. This method verifies that a user's UID from the identity provider ends with their registered national ID. If a match is found, it securely associates the user without altering their session. If there is no match, or if the national ID is absent, the method triggers a logout and redirects the user to the registration page, ensuring only properly identified users are associated.

Key improvements include:

1. Ensuring accurate user associations by verifying the UID-national ID suffix match.
2. Automatically handling sessions by logging out mismatched users to enhance security.
3. Streamlining user onboarding with an immediate redirect to the registration page if the national ID is not present.

## Testing instructions
This requires the following settings

```python
REGISTRATION_EXTENSION_FORM_FIELDS_TPA_OVERRIDES = ["arabic_name", "national_id"]
REGISTRATION_EXTRA_FIELDS = {
    ...
    "national_id": "required",
    ...
}
```
Replace `"social_core.pipeline.social_auth.associate_user"` with `"eox_nelp.third_party_auth.pipeline.validate_national_id_and_associate_user",` in `SOCIAL_AUTH_TPA_SAML_PIPELINE`

image
image

Finally register a new user with SAML

### DIFF
--- a/eox_nelp/third_party_auth/pipeline.py
+++ b/eox_nelp/third_party_auth/pipeline.py
@@ -4,6 +4,8 @@ functions:
     social_details: Allows to map response fields to user standard fields.
     invalidate_current_user: Sets to None the current user.
 """
+import logging
+
 from django.conf import settings
 from django.contrib.auth import logout
 from django.http import HttpResponseForbidden
@@ -14,6 +16,8 @@ from social_core.pipeline.social_auth import social_details as social_core_detai
 
 from eox_nelp.edxapp_wrapper.edxmako import edxmako
 from eox_nelp.third_party_auth.utils import match_user_using_uid_query
+
+logger = logging.getLogger(__name__)
 
 
 def social_details(backend, details, response, *args, **kwargs):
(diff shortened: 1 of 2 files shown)


eox_nelp/third_party_auth/pipeline.py
```diff
@@ -143,3 +145,30 @@ def disallow_staff_superuser_users(  # pylint: disable=unused-argument
...
+        Otherwise, logs out the current session and redirects to the registration page.
+    """
+    national_id = user.extrainfo.national_id if user and hasattr(user, "extrainfo") else ""
+
+    if national_id and uid.endswith(national_id):
+        return associate_user(backend, uid, user, social, *args, **kwargs)
+
+    logout(request)
+
+    return redirect("/register")
```
johanseto, line 174: Here is a redirect. Do you think this pipe would have `@partial `decorator? To continue the process in this step pipe???
https://python-social-auth.readthedocs.io/en/latest/pipeline.html#partial-pipeline

andrey-canon: That decorator doesn't fit in this context at all,  when I was developing this pipe I included this `backend.strategy.clean_partial_pipeline(partial_token)` to restart the current partial then I realized that is not necessary because `logout(request)` performs that operation, basically this pipe aims to restart the whole pipeline when the national id condition is not valid in order to start a new process so doesn't make sense to start  a new process from this step

johanseto (approved):

PR #234, opened 2024-12-10
fix: templates loading reset

johanseto:
## Description

This fix the order ready method of eox_nelp works. Mainly this avoid that if eox_nelp is intalled before eox_theming. eox_theming should not reset look_ups in it ready method. https://github.com/eduNEXT/eox-theming/blob/v8.1.0/eox_theming/apps.py#L58

So with this change eox_nelp exteding template in the ready method would work. https://github.com/eduNEXT/eox-nelp/blob/8c8306e59e94bc0209c49d044dc5d5c12874be98/eox_nelp/init_pipeline.py#L60

## Testing instructions
Go to stats view in redwood dev kit or in an image where eox_nelp was installed before eox_theming.

image
`http://local.overhang.io:8000/eox-nelp/stats/tenant/`

### Before

Screenshot from 2024-12-10 14-37-43
Screenshot from 2024-12-10 14-37-50
Screenshot from 2024-12-10 14-46-02

### After

Screenshot from 2024-12-10 14-38-48
Screenshot from 2024-12-10 14-38-05
Screenshot from 2024-12-10 14-46-20
## Additional information

I tried EDNX_INSTALLED_APPS spec but doesn't work.
Screenshot from 2024-12-10 14-36-39

## Checklist for Merge

- [ ] Updated documentation
- [ ] Rebased master/main
- [ ] Squashed commits

### DIFF
--- a/eox_nelp/settings/common.py
+++ b/eox_nelp/settings/common.py
@@ -26,6 +26,7 @@ DEFAULT_FUTUREX_NOTIFY_SUBSECTION_SUBJECT_MESSAGE = (
 
 JSON_API_REST_FRAMEWORK = 'rest_framework_json_api'
 EOX_AUDIT_MODEL_APP = 'eox_audit_model.apps.EoxAuditModelConfig'
+EOX_NELP_MODEL_APP = 'eox_nelp.apps.EoxNelpConfig'
 
 
 def plugin_settings(settings):
@@ -67,7 +68,6 @@ def plugin_settings(settings):
         settings.INSTALLED_APPS.append(JSON_API_REST_FRAMEWORK)
     if find_spec('eox_audit_model') and EOX_AUDIT_MODEL_APP not in settings.INSTALLED_APPS:
         settings.INSTALLED_APPS.append(EOX_AUDIT_MODEL_APP)
-
     try:
         payments_notifications_context = (
             'eox_nelp.payment_notifications.context_processor.payments_notifications_context'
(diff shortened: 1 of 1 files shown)


andrey-canon:
> I tried EDNX_INSTALLED_APPS spec but doesn't work.

Could you try again ? I tried that solution on my local machine and that worked

eox_nelp/settings/common.py
```diff
@@ -82,3 +82,7 @@ def plugin_settings(settings):
     except AttributeError:
         # We must find a way to register this error
         pass
+    # Ensure EOX_NELP is the last plugin installed app for ready run.
+    if EOX_NELP_MODEL_APP in settings.INSTALLED_APPS and settings.INSTALLED_APPS[-1] != EOX_NELP_MODEL_APP:
```
andrey-canon, line 86: could you move this logic to the settings file(nelc-dev-kit/plugins/extra_env_patches.py and  manifest-sagar/plugins) ?? I make this suggestion because you're fixing the app position and if we need to change that later, this implementation makes harder to modify that order

johanseto: > could you move this logic to the settings file(nelc-dev-kit/plugins/extra_env_patches.py and manifest-sagar/plugins) ?? I make this suggestion because you're fixing the app position and if we need to change that later, this implementation makes harder to modify that order

Yes I could, also in the plugin you could reorder. As you see custom_reg_form is appended using the env patches. Anyway, if you think is better the fix there I could do it.

johanseto:
> > I tried EDNX_INSTALLED_APPS spec but doesn't work.
> 
> Could you try again ? I tried that solution on my local machine and that worked

@andrey-canon I am trying it in this way. But is not working, is something I am doing wrong?
image

Maybe only work in tenant_config?

andrey-canon:
> > > I tried EDNX_INSTALLED_APPS spec but doesn't work.
> > 
> > 
> > Could you try again ? I tried that solution on my local machine and that worked
> 
> @andrey-canon I am trying it in this way. But is not working, is something I am doing wrong? image
> 
> Maybe only work in tenant_config?

I tested this with tenant config, but I think this should work with general settings if the tenant has the EDNX_USE_SIGNAL setting

johanseto:
Heeding the recommendations of the PR, this fix would be handled by env patches settings.
https://github.com/nelc/nelc-dev-kit/pull/86
https://github.com/nelc/manifest-sagar/pull/654

PR #238, opened 2025-02-11
chore: remove external clients api package

Ang-m4:
## Description
This PR removes the deprecated `api_clients package`, which has been replaced by the updated `external_api_clients` package. The changes include revising and sorting import statements and deleting the outdated package.

### SMS Vendors
I sent an SMS to check if the service was available. Although I didn't receive the expected response, the 401 status code is enough proof that it's working.

sms_vendor_test_eox_nelp


Ang-m4:
Since the changes in this PR exceed the recommended size of 1000 lines, I decided to split it into several smaller PRs that will be submitted soon. Additionally, I noticed an issue with the person_engine client package that needed to be addressed first.

PR #241, opened 2025-02-13
feat: upgrade python version and django version

johanseto:
# Description

Run tests with django version 4.0.

These are the python and Django versions that our images are using.

Screenshot from 2025-02-13 12-47-13

## Before

Screenshot from 2025-02-13 14-45-47


Ang-m4 (approved):
LGTM

PR #244, opened 2025-02-14
fix: update pearson rti api client package

Ang-m4:
## Description
This PR removes the unused `pearson_rti` module from the deprecated `api_clients` package.

## Changes
I just deleted the unused modules and test, and change the necessary imports to the updated `external_api_clients` package.

## Testing
As this module was removed from the new `external_api_clients` and the `pearson_vue` package is going to be refactored soon. Tests are not needed.  For any `CDD`, `EAD` or `RTI` tests please check the PR #245.


andrey-canon (approved):
This looks ok since this functionality must removed from this repository

eox_nelp/pearson_vue/utils.py
```diff
@@ -67,3 +68,88 @@ def update_user_engines(user, action_type, course_id=None):
...
+
+    Returns:
+        dict: The exam data formatted for the request.
+    """
+    return {
+        "external_key": exam_id,
+    }
+
+
+def filter_action_parameters(action_name, user, exam_id=None):
```
johanseto, line 128: At the end this would be called a filter?

eox_nelp/pearson_vue/utils.py
```diff
@@ -67,3 +68,88 @@ def update_user_engines(user, action_type, course_id=None):
...
+        "exam_data": get_exam_data(exam_id),
+    }
+
+    if action_name == "cdd":
+        del action_parameters["exam_data"]
+        return action_parameters
+
+    if action_name == "ead":
+        del action_parameters["platform_data"]
+        return action_parameters
```
johanseto, line 153: Finally. it was necessary to remove those kwargs?
```suggestion
```

Ang-m4: At the end it was, but that change is going to be applied at #245.
For now, this PR only includes the `pearson_rti` client removal. 
I just updated this PR removing unnecessary commits that were causing this changes to be shown.

PR #248, opened 2025-02-14
fix: update sms vendor api client package

Ang-m4:
## Description
I removed the unused `sms_vendor` module from the deprecated `api_clients` package.  imports were updated.

## Testing
I sent an SMS to check if the service was available. Although I didn't receive the expected response, the 401 status code is enough proof that it's working.

sms_vendor_test_eox_nelp


eox_nelp/signals/tasks.py
```diff
@@ -61,7 +61,7 @@ def _post_futurex_progress(data):
         data (dict): dict to send to futurex enrollment-progress path.
     """
     api_client = FuturexApiClient()
-    response = api_client.enrollment_progress(data)
+    response = api_client.enrollment_progress(data)  # pylint: disable=no-member
```
andrey-canon, line 64: Why are you changing this line ?

Ang-m4: That line was changed because at some cases the GitHub checks as well as the local quality tests kept failing showing the Pylint no-member error. 
Seems now that its not necessary.

andrey-canon: I have tested the quality tests  and if I remove this changes the result is the same, please revert this

Ang-m4: Got it!

andrey-canon (approved):
LGTM

PR #250, opened 2025-02-21
feat: add required fields validation logic

andrey-canon:
### Description
This PR adds a method for validating user fields based on the REQUIRED_USER_FIELDS setting.
The validation ensures that each field meets the specified constraints, such as max_length, char_type, format, and optional_values.

### Changes Introduced
Implemented validate_required_user_fields, which checks user fields against predefined validation rules.
Added specific validation methods for max_length, char_type, format, and optional_values.
Integrated structured settings to define required fields.
Added unit tests to ensure correct validation behavior.
### How to Test
1. Ensure the following setting is configured 
```
REQUIRED_USER_FIELDS = {
    "account": {
        "first_name": {"max_length": 30, "char_type": "latin"},
        "last_name": {"max_length": 50, "char_type": "latin"},
    },
    "profile": {
        "city": {"max_length": 32, "char_type": "latin"},
        "country": {"max_length": 2, "optional_values": ["US", "CA", "MX", "BR"]},
        "phone_number": {"max_length": 15, "format": "phone"},
        "mailing_address": {"max_length": 40},
    },
    "extra_info": {
        "arabic_name": {"max_length": 20, "char_type": "arabic"},
        "arabic_first_name": {"max_length": 20, "char_type": "arabic"},
        "arabic_last_name": {"max_length": 50, "char_type": "arabic"},
    },
}
```

2. Open a Django shell session:
` ./manage.py lms shell`
3. Run the validation manually:
```
from eox_nelp.user_profile.required_fields_validation import validate_required_user_fields
from django.contrib.auth import get_user_model

User = get_user_model()
user = User.objects.get(username="your username")

validate_required_user_fields(user)
```
4. Modify the user fields in the Admin Panel

- Change values to invalid inputs (e.g., exceeding max_length, using incorrect char_type, or entering an invalid format).
- Run the validation again and verify that invalid fields are correctly identified.
### Expected Outcome
Fields that do not comply with the settings should be flagged as invalid.
The function should return a dictionary containing invalid fields and their respective errors.

PR #251, opened 2025-02-21
fix: reqs contraints absolute -> relative

johanseto:
# Description

basically I ran make upgrade with pip==24.2

I was using pip 25.0.1, and the constraints are saved absolute and not relative.


johanseto:
Closed in favor of https://github.com/eduNEXT/eox-nelp/pull/250

PR #258, opened 2025-03-05
feat: modify ExtraInfo saving to support multiple fields and enforce …

andrey-canon:
## Description

Refactored the save_extrainfo function to support updating multiple ExtraInfo fields at once.

Previously, the function only allowed updating a single field per call and directly saved the data without validation.

Now, it uses a ModelForm to ensure that all updates pass validation before saving, improving data integrity and consistency.

## How to test
1. Go to `/eox-nelp/api/user-profile/v1/update-user-data/`
2. Set the setting `USER_PROFILE_API_EXTRA_INFO_FIELDS`
```python
USER_PROFILE_API_EXTRA_INFO_FIELDS = [
    "arabic_first_name",
    "arabic_name",
    "arabic_last_name",
    "national_id",
]
```
3. Make multiple request with valid and invalid data, also the user may or may not have extra info

##### Data examples
```json
{
    "arabic_first_name": "مسؤل",
    "arabic_name": "مسؤل مسؤل",
    "arabic_last_name": "مسؤل",
    "national_id": "12345678",
}

{
    "arabic_first_name": "مسؤل",
    "arabic_name": "invalid",
    "arabic_last_name": "مسؤل",
    "national_id": "12345678",
}

{
    "arabic_first_name": "مسؤل",
    "arabic_name": "invalid",
    "arabic_last_name": "مسؤل",
    "national_id": "invalid",
}
```

### DIFF
--- a/eox_nelp/one_time_password/view_decorators.py
+++ b/eox_nelp/one_time_password/view_decorators.py
@@ -12,7 +12,7 @@ from django.core.cache import cache
 from django.http import HttpResponseForbidden, JsonResponse
 from rest_framework import status
 
-from eox_nelp.utils import save_extrainfo_field
+from eox_nelp.utils import save_extrainfo
 
 logger = logging.getLogger(__name__)
 
@@ -57,7 +57,7 @@ def validate_otp(func):
             if not proposed_user_otp == cache.get(user_otp_key):
                 return HttpResponseForbidden(reason="Forbidden - wrong code")
 
-            save_extrainfo_field(request.user, "is_phone_validated", True)
+            save_extrainfo(request.user, {"is_phone_validated": True})
             cache.delete(user_otp_key)
 
         return func(request)
--- a/eox_nelp/user_profile/api/v1/views.py
+++ b/eox_nelp/user_profile/api/v1/views.py
@@ -23,7 +23,7 @@ from eox_nelp.edxapp_wrapper.user_api import accounts, errors
 from eox_nelp.one_time_password.view_decorators import validate_otp
 from eox_nelp.pearson_vue_engine.tasks import real_time_import_task_v2
 from eox_nelp.user_profile.required_fields_validation import validate_required_user_fields
-from eox_nelp.utils import save_extrainfo_field
+from eox_nelp.utils import save_extrainfo
 
 logger = logging.getLogger(__name__)
 
@@ -71,10 +71,12 @@ def update_user_data(request):
             # Also some fields related ExtraInfo are not editable too  in the standard implementation. So we need
             # save_extrainfo_field method with the desired settings.
             required_user_extra_info_fields = getattr(settings, 'USER_PROFILE_API_EXTRA_INFO_FIELDS', [])
+            extra_info_data = {
+                field: request.data[field] for field in required_user_extra_info_fields if field in request.data
+            }
 
-            for field in required_user_extra_info_fields:
-                if value := request.data.get(field):
-                    save_extrainfo_field(request.user, field, value)
+            if extra_info_data:
+                save_extrainfo(request.user, extra_info_data)
 
     except errors.AccountValidationError as err:
         return Response({"field_errors": err.field_errors}, status=status.HTTP_400_BAD_REQUEST)
--- a/eox_nelp/utils.py
+++ b/eox_nelp/utils.py
@@ -2,6 +2,9 @@
 import re
 from copy import copy
 
+from custom_reg_form.forms import ExtraInfoForm
+from custom_reg_form.models import ExtraInfo
+from django.core.exceptions import ValidationError
 from opaque_keys.edx.keys import CourseKey
 
 from eox_nelp.edxapp_wrapper.course_overviews import get_course_overviews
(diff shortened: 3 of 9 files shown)


eox_nelp/one_time_password/view_decorators.py
```diff
@@ -57,7 +57,7 @@ def wrapper(request):
             if not proposed_user_otp == cache.get(user_otp_key):
                 return HttpResponseForbidden(reason="Forbidden - wrong code")
 
-            save_extrainfo_field(request.user, "is_phone_validated", True)
+            save_extrainfo(request.user, {"is_phone_validated": True})
```
johanseto, line 60: This works, even though the form doesnt have "is_phone_validated" field.
https://github.com/nelc/nelp-custom-registration-fields/blob/714b145/custom_reg_form/forms.py#L17
Django powers

eox_nelp/utils.py
```diff
@@ -156,21 +159,26 @@ def camel_to_snake(string):
...
+    form_data = extra_info.__dict__
+    form_data.update(validated_data)
+
+    form = ExtraInfoForm(form_data, instance=extra_info)
+
+    if form.is_valid():
+        form.save()
     else:
-        ExtraInfo.objects.create(user=user, **{field: value})  # pylint: disable=no-member
+        raise ValidationError(form.errors)
```
johanseto, line 184: Using the form as a method to update extraInfo might seem good, but I have some concerns about potential risks.
 Specifically, for users who already exist in the database but bypassed the form, there could be inconsistencies. While these users are well-formed in terms of the model and database (and can even be loaded in the admin panel), they may not meet the form's validation requirements. For example, the national_id field has blank=False in the form, but many users might have `national_id` set to NULL in the database.

This creates a problem when attempting to update a specific field in extraInfo, such as is_phone_validated. In such cases, the form validation would fail due to the national_id constraint, even if we only intend to update `is_phone_validated`. This limitation could hinder the ability to perform partial updates efficiently.

[Screencast from 06-03-25 15:23:00.webm](https://github.com/user-attachments/assets/ec4f2dae-1eb3-4cee-806b-95419e1757ad)

I think this error could also happen with new fields, like sector that are blank=False and Null=True.

andrey-canon: basically that is the proposal, any user won't be able to update any field if they don't have the whole data so you have to pass all missing fields otherwise the API should return bad request

eox_nelp/user_profile/api/v1/views.py
```diff
@@ -71,10 +71,12 @@ def update_user_data(request):
...
+            }
 
-            for field in required_user_extra_info_fields:
-                if value := request.data.get(field):
-                    save_extrainfo_field(request.user, field, value)
+            if extra_info_data:
+                save_extrainfo(request.user, extra_info_data)
 
     except errors.AccountValidationError as err:
         return Response({"field_errors": err.field_errors}, status=status.HTTP_400_BAD_REQUEST)
```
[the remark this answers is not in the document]
andrey-canon: Good catch, it was hard to replicate this error since all my fields were ok and I had to modify that via shell

eox_nelp/one_time_password/view_decorators.py
andrey-canon: Indeed it's not working, the test is wrong and it's marking a false positive

andrey-canon: I added that to this pr https://github.com/nelc/nelp-custom-registration-fields/pull/14

eox_nelp/user_profile/api/v1/views.py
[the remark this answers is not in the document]
andrey-canon: Fixed

eox_nelp/utils.py
```diff
@@ -156,21 +159,26 @@ def camel_to_snake(string):
...
+    validated_data = {field: value for field, value in data.items() if hasattr(ExtraInfo, field)}
+
+    if not validated_data:
         return
 
-    if extra_info := getattr(user, "extrainfo", None):
-        setattr(extra_info, field, value)
-        extra_info.save()
+    extra_info, _ = ExtraInfo.objects.get_or_create(user=user)  # pylint: disable=no-member
+    form_data = extra_info.__dict__
```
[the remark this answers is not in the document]
andrey-canon: implemented

eox_nelp/one_time_password/view_decorators.py
andrey-canon: fixed

johanseto (approved):
Remember rebase before merging due few test conflicts.

Anyway, the new change seems working as expected.

- is_phone_validated saving

[Screencast from 07-03-25 11:23:39.webm](https://github.com/user-attachments/assets/3805650b-9086-4187-a138-96b3542cb832)

- the API with absence of national_id returns badrequest 

[Screencast from 07-03-25 11:30:08.webm](https://github.com/user-attachments/assets/566c745d-a679-4133-b193-6325ee314e37)

PR #259, opened 2025-03-06
feat: add sqs client requirement

johanseto:
## Description

feat: add sqs client requirement

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -16,12 +16,18 @@ asgiref==3.8.1
     # via
     #   django
     #   django-simple-history
-attrs==25.1.0
+attrs==25.3.0
     # via openedx-events
 beautifulsoup4==4.13.3
     # via bs4
 billiard==4.2.1
     # via celery
+boto3==1.37.13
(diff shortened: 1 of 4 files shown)


requirements/base.txt
```diff
@@ -275,7 +288,7 @@ path==16.16.0
     # via edx-i18n-tools
 pbr==6.1.1
     # via stevedore
-phonenumbers==8.13.55
+phonenumbers==9.0.0
```
andrey-canon, line 291: could you check if this change breaks something ?

andrey-canon (changes requested):
Looks ok however could you rebase to branch  to remove tox changes

requirements/base.txt
johanseto: Seems that in python nothing. Anyway, we need to keep monitoring in case.
https://github.com/daviddrysdale/python-phonenumbers/blob/dev/python/HISTORY.md?plain=1#L15-L19

andrey-canon (approved):
LGTM

PR #273, opened 2025-05-12
feat: update report reasons to use 'IC' for inappropriate content

Ang-m4:
## Description
This pull request updates the `reason` field in the `ReportCourse` and `ReportUnit` models to replace the code `SC` (Sexual Content) with `IC` (Inappropriate Content). It includes changes to the model definition, a migration to alter the field, and a data migration to update existing records in the database.

### Model Updates:
* Updated the `reason` field in the `BaseReport` model to use `IC` ("Inappropriate content") instead of `SC` ("Sexual content"). This change applies to both the field's description and its choices.

### Schema Migration:
* Added a migration (`0015_alter_reportcourse_reason`) to update the `reason` field in the `ReportCourse` model. The field now includes `IC` as a valid choice and removes `SC`.

### Data Migration:
* Added a migration (`0016_data_update_sc_to_ic_report_reason`) to update all existing database records where the `reason` field is set to `SC` to now use `IC`. This migration also includes a reverse operation to revert the changes if needed. 

## How to test

1. Go to your lms container and access a bash terminal
```bash
docker exec -it tutor_dev-lms-1 bash
```

2. Run the eox_nelp migrations
```bash
./manage.py lms migrate eox_nelp
```
image

### Before migrations
image

### After migrations:
image


johanseto (commented):
Why do we have 2 migrations for this requirement. Is not possible to have only once??

Ang-m4:
> Why do we have 2 migrations for this requirement. Is not possible to have only once??

I think is better to separate the schema change and definition (0015), automatically created using `makemigrations`.
From the data update migration (0016) - Custom migration.

If you want to see them both merged in a single custom migration, please let me know.

eox_nelp/migrations/0016_data_update_sc_to_ic_report_reason.py
```diff
@@ -0,0 +1,40 @@
+# Generated by Django 4.0.10 on <current_date>
```
johanseto, line 1: So this is a custom migration. With this line look like auto xD

johanseto (approved):
LGTM and working in my local env.




Let's merge it, and we should deploy it using master with this last change.

PR #283, opened 2025-07-29
feat: log update_mt_training_stage response and verify it in test

andrey-canon:
## Description

Adds a new log line to update_mt_training_stage to include the full API response.
Also updates the corresponding unit test to assert the exact INFO log output.

This ensures traceability of the response returned by the external service and improves confidence through explicit log validation.


johanseto (approved):
:shipit:

PR #285, opened 2025-07-31
fix: this adds a missing migration that was not generated earlier

andrey-canon:
## Description

This migration should have been included in this [pr](https://github.com/nelc/eox-nelp/pull/273) however that was not the case so this pr adds it


johanseto (approved):
Working in my local env.
Screenshot from 2025-08-01 11-07-07

PR #304, opened 2026-03-19
feat: add Cloud CDN signed cookie middleware and unit tests

andrey-canon:
## Summary

This PR introduces a middleware to generate and attach Google Cloud CDN signed cookies, enabling secure access to private assets stored in GCS.

## Changes

- Added `GCPCloudCDNSignedCookieMiddleware`
  - Generates signed cookies for authenticated users
  - Uses Cloud CDN signing key configured in the backend bucket
  - Manually sets the `Set-Cookie` header to avoid quoting issues
- Added generic settings:
  - `GCP_CLOUD_CDN_SIGNING_KEY_NAME`
  - `GCP_CLOUD_CDN_SIGNING_KEY`
  - `GCP_CLOUD_CDN_URL_PREFIX`
  - `GCP_CLOUD_CDN_COOKIE_MAX_AGE`
  - `GCP_CLOUD_CDN_COOKIE_DOMAIN`
- Added unit tests covering:
  - Cookie generation
  - Signature generation
  - Authenticated vs unauthenticated behavior
  - Cookie domain handling

## Motivation

Cloud CDN requires signed cookies (or signed URLs) to serve content from private GCS buckets.

Since Open edX generates asset URLs with query parameters (e.g. `?v=...`), signed URLs are not viable.  
This implementation enables access using signed cookies without modifying existing URL generation logic.

## Testing

- Unit tests implemented and passing
- Manual validation performed using browser and `curl`
- Verified:
  - Cookie format is valid for Cloud CDN
  - Cookie must not include quotes
  - Cookie name must be `Cloud-CDN-Cookie`
  - Access works correctly when cookie is present

## Notes

- Requires a configured Cloud CDN backend bucket with a signed URL key
- Signing key must match Terraform configuration
- Middleware applies globally and can be reused for any CDN-protected assets

### DIFF
--- a/eox_nelp/middleware.py
+++ b/eox_nelp/middleware.py
@@ -1,4 +1,109 @@
 """Middleware file.
 
-Required NELP middlewares that allow to customize edx-platform.
+Required NELP middlewares that allow customizing edx-platform.
 """
+
+import base64
+import hashlib
+import hmac
+import time
+
+from django.conf import settings
+
+
+class GCPCloudCDNSignedCookieMiddleware:
+    """
+    Middleware that generates and attaches a Google Cloud CDN signed cookie.
+
+    The cookie is only generated for authenticated users and is signed using
+    the key configured for the Cloud CDN backend bucket.
+    """
+
+    COOKIE_NAME = "Cloud-CDN-Cookie"
+
+    def __init__(self, get_response):
+        """
+        Initialize middleware with Django's response handler.
+        """
+        self.get_response = get_response
+
+    def _sign(self, data: bytes) -> str:
+        """
+        Generate the HMAC-SHA1 signature required by Google Cloud CDN.
+
+        - Decodes the base64url key from settings.
+        - Signs the payload using HMAC-SHA1.
+        - Returns a base64url-encoded signature without padding.
+
+        Args:
+            data (bytes): The payload to sign.
+
+        Returns:
+            str: Base64url-encoded signature.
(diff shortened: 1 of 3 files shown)


eox_nelp/middleware.py
```diff
@@ -1,4 +1,109 @@
...
+    def __call__(self, request):
+        """
+        Process each request and attach the signed cookie to the response
+        if the user is authenticated.
+
+        The cookie is manually constructed to avoid Django adding quotes,
+        which would break Cloud CDN validation.
+        """
+        response = self.get_response(request)
+
```
johanseto, line 92: Could we use here a setting to activate or activate this middleware validation and built cookie logic. Just like:
```python3
if not getattrr(settings, "ACTIVATE_GCP_CLOUD_CDN_COOKIE, None):
    return response
```

andrey-canon: Is there a specific reason to keep the middleware? If we aren't generating the cookie, it might be easier to just pull the middleware out of the settings entirely.

johanseto: Yes, but for the tenant config, it is not easier to change and set all the middlewares.

andrey-canon: IMO, we will never encounter that specific scenario, and adding this would only introduce unnecessary complexity to the configuration layer.

johanseto (approved):

PR #310, opened 2026-05-22
perf(api): add 5-minute cache to mfe config endpoint

andrey-canon:
## Description
This PR introduces a caching layer to the `eox-nelp` MFE configuration API endpoint to optimize backend response times and eliminate redundant Python processing overhead.

Since the MFE frontend already caches this configuration locally for 5 minutes, aligning the backend view with a 300-second TTL prevents unnecessary CPU and serialization load without affecting the end-user experience.

## Changes
* Imported and applied `@method_decorator(cache_page(...))` to the `get` method in `mfe_config_api/v1/views.py`.
* Added the `EOX_NELP_MFE_CONFIG_API_CACHE_TIMEOUT = 300` setting to `common.py` and `test.py` to control the TTL centrally.

## Impact
* **Performance:** The MFE configuration response is now cached for 5 minutes, significantly reducing CPU processing, dictionary merging, and JSON serialization overhead during high-traffic login and load events.
* **Maintainability:** Relies on Django's native caching utilities, respecting current multitenant isolation out of the box.

### DIFF
--- a/eox_nelp/mfe_config_api/v1/views.py
+++ b/eox_nelp/mfe_config_api/v1/views.py
@@ -4,6 +4,8 @@ import json
 
 from django.conf import settings
 from django.http import JsonResponse
+from django.utils.decorators import method_decorator
+from django.views.decorators.cache import cache_page
 from eox_theming.configuration import ThemingConfiguration as theming
 from rest_framework import status
 
@@ -45,7 +47,7 @@ class NelpMFEConfigView(MFEConfigView):
     }
     ```
     """
-
+    @method_decorator(cache_page(settings.EOX_NELP_MFE_CONFIG_API_CACHE_TIMEOUT))
     def get(self, request):
         """
         Return the MFE configuration by NELP(adding custom nelp fields).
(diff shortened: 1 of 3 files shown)


eox_nelp/mfe_config_api/v1/views.py
```diff
@@ -45,7 +47,7 @@ class NelpMFEConfigView(MFEConfigView):
     }
     ```
     """
-
+    @method_decorator(cache_page(settings.EOX_NELP_MFE_CONFIG_API_CACHE_TIMEOUT))
```
johanseto, line 50: why you choose to use method decorator? instead of directly `@cache_page` ? to use more than one decorator in future?

andrey-canon: You must use method_decorator because Django's standard view decorators (like @cache_page) are built for function-based views (FBVs). They expect a request object as their very first argument. The method_decorator tool acts as a bridge. It converts a function decorator into a method decorator. https://docs.djangoproject.com/en/6.0/topics/class-based-views/intro/#decorating-the-class

johanseto (approved):

PR #311, opened 2026-05-28
feat: add course_experience indexes to improve get

johanseto:
This could help the get_or_create methods. Mainly the get step.

## Description

Based in the reason why the courseware student module is not having these problems of insert or updates being called more times.

https://github.com/openedx/openedx-platform/blob/b1b59ddc5a7d8af31091c74d1f3b75ff38acde3e/lms/djangoapps/courseware/models.py#L102-L106

image

### test

Migrated in my local env

image

## Checklist for Merge

- [ ] Updated documentation
- [ ] Rebased master/main
- [ ] Squashed commits


andrey-canon (commented):
Thanks for the reference and the effort to optimize this. However, there is a fundamental difference between the Open edX core code you linked and the implementation in this PR.

In the Open edX core link:
* `unique_together` creates a unique index on: `('student', 'module_state_key', 'course_id')`.
* The `models.Index` creates a completely different, non-unique composite index on: `['module_state_key', 'grade', 'student']` (different fields, different order) specifically designed to optimize grading queries (`courseware_stats`). This is not redundant.

In our current PR implementation:
```python
unique_together = [["author", "item_id"]]
indexes = [
models.Index(fields=["author", "item_id"], name="likedislike_author_item_idx"),
]
```
Here, both blocks point to the exact same fields (`author` and `item_id`) in the exact same order. Database engines automatically generate an underlying composite index to enforce `unique_together`. Adding an explicit `models.Index` on top of it creates a duplicate, redundant index over the same columns. This wastes disk space and degrades write performance (`INSERT`/`UPDATE`) because the database has to update two identical index trees.

johanseto:
Thanks for the excellent breakdown! You are completely right that in the current PR implementation, the models.Index is 100% redundant and creates a duplicate index tree because the database already builds a unique index for unique_together on ('author', 'item_id'). I appreciate you catching that.

However, there is a distinct performance reason to keep a composite index here, but with a crucial twist: we need to change the order of the fields.

1. The Field Order Difference
If we change the explicit index to look like this:

Python


indexes = [
    models.Index(fields=["item_id", "author"], name="likedislike_item_author_idx"),
]
It is no longer redundant. Databases read composite indexes from left to right.

The unique_together index optimizes queries filtering by author or author + item_id.

This new index optimizes queries filtering by item_id alone, which is incredibly common when fetching all likes/dislikes for a specific piece of content.

2. The Django get_or_create Problem
More importantly, this specific composite index helps mitigate a well-known performance bottleneck with Django’s get_or_create().

When Django executes get_or_create, it first attempts a GET query. If it misses, it handles the exception and issues an INSERT. In high-concurrency environments or on tables with heavy traffic, that initial GET lookup needs to be as fast as possible to prevent thread blocking and DB lock contention.

By having optimized indexes covering both lookup angles (author first and item_id first), we ensure that Django's internal GET check is lightning-fast, significantly reducing the performance degradation typically associated with get_or_create.

Moving Forward
I propose we keep an explicit index but flip the column order to ["item_id", "author"]. This eliminates the redundancy you pointed out, speeds up content-level lookups, and protects our get_or_create calls.

Let me know what you think!



And also adding the index for get, could potentially penalize the update, but actually the cpu of the update transation of like and dislike is really slow. The problem is with the lock. 

image

andrey-canon:
> When Django executes get_or_create, it first attempts a GET query. If it misses, it handles the exception and issues an INSERT. In high-concurrency environments or on tables with heavy traffic, that initial GET lookup needs to be as fast as possible to prevent thread blocking and DB lock contention.

Thanks for the follow-up. To be fair, changing the column order to `["item_id", "author"]` **could indeed improve the performance of specific standalone SELECT queries**—specifically those that filter strictly by `item_id` (like fetching all likes for a single piece of content). 

However, this minor read optimization comes at a direct and heavy cost to write performance. 

Every single new index adds an operational penalty to `INSERT` and `UPDATE` statements. In high-concurrency tables like likes/dislikes, adding a reversed index forces the database engine to maintain and update two separate index trees for every single event. This makes the write transactions heavier and slower, directly increasing the duration of the row/gap locks and exacerbating the exact locking issues that cause bottlenecks during concurrent insertions.

Furthermore, standard `SELECT` queries are non-blocking read operations that never cause database lock contention or thread blocking for other transactions, meaning the initial `GET` step is not the source of the platform's lock issues.

Based on this technical trade-off, I don't consider these changes to be relevant or necessary for our current scale. Since the primary lookup is already fully covered by the existing constraint and we want to avoid introducing extra write overhead to the execution phase that actually handles the locks, it might be better to hold off on this for now to keep the database schema lean. We can definitely revisit this in the future if  metrics show that standalone item_id lookups are becoming a real bottleneck. Let me know your thoughts on this approach.

johanseto:
Discarded in favour of https://github.com/nelc/eox-nelp/pull/314

PR #312, opened 2026-05-28
Jlc/redis writting proxy course experience

johanseto:
### Summary
- Introduces a cache-backed proxy path in the ExperienceView that, when enabled, serves reads from cache and writes to cache (with asynchronous persistence to DB).
- Aims to improve responsiveness for read-after-write scenarios by returning an in-memory instance immediately while persisting changes to the database via a Celery task.

#### What changed
- views.py
  - get_object: if EOX_NELP_EXPERIENCE_CACHE_ENABLED is True, attempt to read the cached entry (get_experience_cache) and return an in-memory model instance (create_memory_instance) instead of hitting the DB.
  - perform_create / perform_update: when cache flag is enabled, short-circuit DB save and call update_create_experience_cache.
  - update_create_experience_cache: merge serializer.validated_data with parser_context kwargs, drop author, stringify values, upsert into cache via upsert_experience_cache, enqueue persist_experience_to_db.delay(...) and set serializer.instance to an in-memory model instance built from cached data.

#### Rationale
- Provide immediate read-after-write behavior for API clients without waiting for synchronous DB writes.
- Offload write-heavy operations to asynchronous workers and reduce load on primary DB in write bursts.

#### Behavioral details & implications
- When the feature flag is enabled:
  - POST/PATCH operations update cache and schedule async persistence; the API returns an in-memory instance that reflects the change immediately.
  - GET operations for a specific resource will return cached data if present.
- When the flag is disabled: existing synchronous DB read/write behavior remains unchanged.
- The change casts all cached values to strings before storing; this may affect downstream consumers expecting typed values.
- Asynchronous persistence is fire-and-forget (persist_experience_to_db.delay). If Celery/worker failures occur, cache changes may not reach DB.

Configuration
- Toggle feature with Django setting: EOX_NELP_EXPERIENCE_CACHE_ENABLED (bool).
- Requires cache backend and Celery workers configured for full functionality.

Backward compatibility & safety
- Default behavior unchanged when the flag is False.
- Serializer.instance is set to an in-memory model instance (not persisted yet) — callers that rely on DB identity should be aware of eventual consistency window.

Testing / QA suggestions
- Unit tests:
  - verify get_object returns memory instance when cache contains entry and flag enabled.
  - verify perform_create/perform_update call update_create_experience_cache and schedule persist_experience_to_db.delay when flag enabled.
  - verify normal DB flow is used when flag disabled.
  - validate handling of missing target_id and that no task is scheduled in that case.
- Integration:
  - End-to-end test with cache + Celery worker verifying eventual DB persistence.
  - Load test to validate DB load reduction and worker throughput.

Potential follow-ups / improvements
- Make cache helper functions (get/set/upsert/delete) themselves respect EOX_NELP_EXPERIENCE_CACHE_ENABLED to centrali

ze flag handling (reduces chances of inconsistent use).
- Avoid casting all values to str; only cast fields that require string form (ids), preserving numeric/boolean types.
- Add error handling/retries/monitoring for persist_experience_to_db tasks to avoid silent data loss.
- Add metrics/tracing around cache hits, misses, queued persistence, and task failures.

#### How to test locally
- Enable the feature in local settings:
  - EOX_NELP_EXPERIENCE_CACHE_ENABLED = True
- Run tests and manual API checks:
  - pytest eox_nelp/course_experience/tests -q
  - Post a item with like and dislike.
  Using cache view it was saved.

View the item, which was also saved in db using django admin.

**Test cache persistence with get**.
Delete the item in django.
The api would return the object, since exists yet in cache.

[Screencast from 28-05-26 15:44:10.webm](https://github.com/user-attachments/assets/dbe1a769-c2cf-4214-a7ee-d2525541e34e)

Files changed
- views.py (primary)

If you want, I can prepare:
- a short unit-test checklist or example tests for the new paths,
- a small patch to make cache helpers honor the feature flag, or
- a rollback/feature-flagging plan to mitigate risks.

**Jira story**: https://edunext.atlassian.net/browse/FUTUREX-1753?atlOrigin=eyJpIjoiMWExZmVmODU3MDlmNDdkYmJmM2Y3YTdjOGJmYzM5OTMiLCJwIjoiaiJ9

### DIFF
--- a/eox_nelp/course_experience/api/v1/tests/mixins_helpers.py
+++ b/eox_nelp/course_experience/api/v1/tests/mixins_helpers.py
@@ -191,6 +191,7 @@ class ExperienceTestMixin:
         expected_data = self.base_data.copy()
         expected_data["data"]["attributes"].update(self.patch_data)
         expected_data["data"]["id"] = None # element by cache does not have id from db.
+
         response = self.client.patch(url_endpoint, self.patch_data, format="multipart")
 
         persist_experience_to_db_mock.delay.assert_called_once()
--- a/eox_nelp/course_experience/api/v1/tests/mixins_helpers.py
+++ b/eox_nelp/course_experience/api/v1/tests/mixins_helpers.py
@@ -182,6 +182,7 @@ class ExperienceTestMixin:
         Test a  patch  request to the detail endpoint for the desired view
         using form data (type document in browser).
         Expected behavior:
+            - persist_experience_to_db task is called once to persist the change to db.
             - Return expected content type.
             - Status code 200.
             - Check the response is equal to the expected patched.
@@ -195,7 +196,6 @@ class ExperienceTestMixin:
         persist_experience_to_db.delay.assert_called_once()
         self.assertIn(response.headers["Content-Type"], RESPONSE_CONTENT_TYPES)
         self.assertEqual(response.status_code, status.HTTP_200_OK)
-        breakpoint()
         self.assertEqual(response.json(), expected_data)
 
     def test_block_url_object_regex(self):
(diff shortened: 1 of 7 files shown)


johanseto:
Deprecated in favour of https://github.com/nelc/eox-nelp/pull/314

PR #314, opened 2026-06-04
feat!: remove db experience foreign key User

johanseto:
## Description

Inspired in courseware student module.
https://github.com/openedx/openedx-platform/blob/b1b59ddc5a7d8af31091c74d1f3b75ff38acde3e/lms/djangoapps/courseware/models.py#L95

The student fields in the User model include this parameter in courseware_studentmodule model, which removes the foreign key constraint in the database. This ensures that if the User table is locked, it won't lock the course_experience table, and vice versa. Similarly, the course_experience table will not block the User table. Additionally, the ORM still handles validations as if the foreign key constraint were present in the code

Comparing the foreign keys in CLoudsql.

### Other models.

Comparing the behavior of inserts, I think this change would be applied primay **completion_blockcompletion**. 
image

But for the moment we could test how it works with eox-nelp.

image

## Checklist for Merge

- [ ] Updated documentation
- [ ] Rebased master/main
- [ ] Squashed commits


andrey-canon (commented):
I like this approach I think this could be very helpful however I think the course_id change is not  necessary I mean  there are almost 1 million of users  however there are only 3500 course overviews IMO that table is not big enough to require this configuration

johanseto:
> I like this approach I think this could be very helpful however I think the course_id change is not necessary I mean there are almost 1 million of users however there are only 3500 course overviews IMO that table is not big enough to require this configuration

I only wanted to make it faster as possible. I mean, in my opinion, the course foreign key is not as important. Also, e.g., course_enrollment doesn't have the foreign key activated too.

https://github.com/openedx/openedx-platform/blob/ff0690f6c8768158dd07ab6a43e88e8695798283/common/djangoapps/student/models/course_enrollment.py#L312

IMO I dont know which advantages could we have having the foreign key of course_overviews.

andrey-canon:
> > I like this approach I think this could be very helpful however I think the course_id change is not necessary I mean there are almost 1 million of users however there are only 3500 course overviews IMO that table is not big enough to require this configuration
> 
> I only wanted to make it faster as possible. I mean, in my opinion, the course foreign key is not as important. Also, e.g., course_enrollment doesn't have the foreign key activated too.
> 
> https://github.com/openedx/openedx-platform/blob/ff0690f6c8768158dd07ab6a43e88e8695798283/common/djangoapps/student/models/course_enrollment.py#L312
> 
> IMO I dont know which advantages could we have having the foreign key of course_overviews.

My main concern is data integrity. The biggest advantage of keeping the CourseOverview foreign key is precisely ensuring that the course actually exists before inserting the record.

If we drop the constraint for author (User), the risk is minimal because the user must be authenticated at the application level to even reach this API; the system already knows the user exists. However, without the course constraint, an authenticated user could easily manipulate the API payload (e.g., via browser console or Postman) and send requests with fake or malformed course IDs. The database would blindly accept them, opening the door to filling the table with inconsistent/junk data, which could lead to 500 errors later when the frontend or reports try to fetch data for ghost courses.

Regarding performance: Does validating against CourseOverview really impact it enough to justify this risk?
Validating an FK against a table with only 3,500 records is practically free. The entire index for that table fits in the RAM (Buffer Pool), making it a microsecond lookup. Furthermore, unlike auth_user (which is constantly being updated and causes severe lock waits), CourseOverview is mostly read-only. There is no real lock contention happening on the course table.

For the author, dropping the constraint gives us a huge performance boost because it bypasses the locks on auth_user. But for the course, the microsecond we save is not worth the risk of data corruption. What do you think about keeping db_constraint=False only for the User and reverting the course one?"

andrey-canon (approved):

johanseto:
## Results

PR #317, opened 2026-07-23
feat(course_modes): automate default mode creation on course creation

andrey-canon:
## Description
This PR introduces an automated, asynchronous flow to generate a default `CourseMode` whenever a new course is created in the platform. This ensures that every new course has a default track (e.g., `honor` or `audit`) configured right out of the box, without requiring manual intervention in the admin panel.

## 🛠️ Technical Implementation
- **Signal Receiver:** Added `receive_course_created_for_modes` which listens to the `COURSE_CREATED` signal from `openedx_events` and triggers a Celery task with a 5-second delay.
- **Celery Task:** Created `create_course_mode` to handle the database operation. It uses `get_or_create` to ensure idempotency and handles missing `CourseOverview` or invalid keys gracefully.
- **Dynamic Configuration:** The default mode slug is dynamically fetched from the `DEFAULT_COURSE_MODE` Django setting (fallback to `"honor"`).
- **Architecture Boundaries:** Created the `CourseMode` backend and wrapper abstractions to comply with the plugin architecture and prevent direct module coupling.
- **Testing:** Added the `CourseMode` model to the `testing_models` migration, created a test backend, and wrote comprehensive unit tests mapping 1:1 with the expected behavior docstrings.

## 🧪 How to Test

1. **Test default behavior:**
   - Log in to **Studio**.
   - Create a new course.
   - Wait a few seconds (to allow the Celery task to execute).
   - Log in to the **Django Admin panel**.
   - Navigate to the **Course Modes** section.
   - **Validation:** Verify that a new `CourseMode` record was automatically created for the course you just made.

2. **Test dynamic settings behavior:**
   - Change your environment's `DEFAULT_COURSE_MODE` setting to a different value (e.g., `"audit"` or `"no-id-professional"`).
   - Restart the LMS/Studio/Celery workers if necessary to pick up the new setting.
   - Go to Studio and create *another* new course.
   - Go back to the **Django Admin panel**.
   - **Validation:** Verify that the newly created course has a `CourseMode` record matching the new setting you configured.


eox_nelp/signals/receivers.py
```diff
@@ -505,3 +506,42 @@ def listen_for_course_delete(sender, course_key, **kwargs):  # pylint: disable=u
...
+    Args:
+        course <CourseData>: CourseData instance containing the course_key.
+            https://github.com/openedx/openedx-events/blob/main/openedx_events/content_authoring/data.py
+        **kwargs: Additional keyword arguments passed with the signal, if any.
+
+    Returns:
+        None: This function does not return any value. It triggers an asynchronous task.
+    """
+    default_mode_slug = getattr(settings, "DEFAULT_COURSE_MODE", "honor")
+
```
johanseto, line 537: How could we disable the feature and use the default course_mode behaviour. I mean if for some reason we need to use the default behaviour of not creating the course_mode, and we could keep the actual way manually.

andrey-canon: There's no way to do that, you could set an invalid course mode to force an error

johanseto, line 537:
```suggestion


if getattr(settings, "DISABLE_DEFAULT_COURSE_MODE_TRIGGER", None):
    return
```
What do you think about adding that

andrey-canon: bad indentation so I can not commit your suggestion

eox_nelp/signals/receivers.py
```diff
@@ -505,3 +506,42 @@ def listen_for_course_delete(sender, course_key, **kwargs):  # pylint: disable=u
...
+    Args:
+        course <CourseData>: CourseData instance containing the course_key.
+            https://github.com/openedx/openedx-events/blob/main/openedx_events/content_authoring/data.py
+        **kwargs: Additional keyword arguments passed with the signal, if any.
+
+    Returns:
+        None: This function does not return any value. It triggers an asynchronous task.
+    """
+    default_mode_slug = getattr(settings, "DEFAULT_COURSE_MODE", "honor")
```
johanseto:
```suggestion
    default_mode_slug = getattr(settings, "DEFAULT_COURSE_MODE", "honor")
    
    if getattr(settings, "DISABLE_DEFAULT_COURSE_MODE_TRIGGER", None):
        return
```

eox_nelp/signals/receivers.py
johanseto: https://github.com/nelc/eox-nelp/pull/317/changes#r3647525974

eox_nelp/signals/receivers.py
```diff
@@ -505,3 +506,45 @@ def listen_for_course_delete(sender, course_key, **kwargs):  # pylint: disable=u
...
+        LOGGER.warning(
+            "receive_course_created_for_modes aborted: valid course data was not provided."
+        )
+        return
+
+    LOGGER.info(
+        "COURSE_CREATED signal received for course %s. Triggering create_course_mode task.",
+        str(course.course_key)
+    )
+
```
johanseto, line 557: Could we add a validation to dont save strange course_modes.

I check the admin, prevent that.

https://github.com/user-attachments/assets/2221450e-3a15-43d8-b44f-c165df087824

https://github.com/openedx/openedx-platform/blob/cc40805a23db7abdece4aa181b132886cd5521dd/common/djangoapps/course_modes/admin.py#L29

Maybe something like

Image

```suggestion
    available_course_enrollment_modes = [mode for mode in settings.COURSE_ENROLLMENT_MODES.keys()]
    if default_mode_slug not in available_course_enrollment_modes:
         LOGGER.warning(
            "receive_course_created_for_modes aborted: valid DEFAULT_COURSE_MODE was not provided"
        )
        return
```

johanseto (commented):
Working in my local env.

https://github.com/user-attachments/assets/a71451d3-8c46-42fd-8390-d6b3fc8efa74

However I suggest a verification to avoid some inconsistent error

https://github.com/user-attachments/assets/6e7a4dd6-493c-4112-8b91-84b0fed0ce03

eox_nelp/signals/receivers.py
andrey-canon: but this means that the user won't know that the configuration is wrong, a regular user doesn't check logs

johanseto: Yes, but the regular instructor user should not configure the `DEFAULT_COURSE_MODE`. it's like to protect that wrong config.

andrey-canon: it doesn't matter, example
User B sets the wrong value 
User A Create a course  and thinks the course mode is ok
Nodoby fixes that, after three moths, someone realizes that is not possible to get the certificate  check course mode table  and fixes that 

on the other hand 
User B sets the wrong value 
User A Create a course  and get an error they complain inmediately and then someone fixes that

andrey-canon: the warning should be enough however our sentry is not reliable

andrey-canon: we could raise an exception instead of printing a warning

johanseto: Maybe not raising an exception to don't stop the flow; I was thinking of a log. error(''). So we could be notified by Sentry

andrey-canon: log error and warning raise the same notification in sentry

johanseto: So the log warning is enough =)

andrey-canon: image

andrey-canon: I thought our current configuration included the warning setting that creates issues, but now I'm not sure. Either way, the comment remains the same regardless of whether the issue is created by a warning or an error.

johanseto (approved):

eox_nelp/signals/receivers.py
andrey-canon: I like this apporach that doesn't allow to create the course if the configuration is wrong  so that forces a good configuration
image

